PR #28, opened 2026-03-11
Fix engine nozzle glow after shutdown and watch button for looped segments

vl3c:
## Summary
- **Bug #18:** Engine nozzle glow now properly clears on shutdown — `EngineShutdown` resets heat animation emissive materials, and `Decoupled`/`Destroyed` events defensively stop all engine/RCS FX before hiding parts
- **Bug #19:** Watch (W) button now works for looped recording segments — pause window path initializes `lastInterpolatedBodyName` from the last trajectory point so `IsGhostOnSameBody` returns true immediately after ghost spawn

## Test plan
- [x] Record a flight with SRB burnout + booster separation, verify ghost nozzle stops glowing after engine cutoff
- [x] Enable loop on a recording, press W — camera should follow the ghost
- [x] Verify W button stays functional across loop cycle boundaries and during pause windows
- [x] Run `dotnet test` — all 1218 tests pass

### DIFF
--- a/Source/Parsek.Tests/SyntheticRecordingTests.cs
+++ b/Source/Parsek.Tests/SyntheticRecordingTests.cs
@@ -507,12 +507,12 @@ namespace Parsek.Tests
             seg0.AddPoint(t + 40, baseLat, baseLon + 0.0079,   160);
             seg0.AddPoint(t + 45, baseLat, baseLon + 0.0084,   110);
             seg0.AddPoint(t + 50, baseLat - 0.0005, baseLon + 0.0088, 80);
-            // Parachute descent — slow final approach
-            seg0.AddPoint(t + 55, baseLat - 0.0008, baseLon + 0.0091, 70);
-            seg0.AddPoint(t + 60, baseLat - 0.0012, baseLon + 0.0093, 62);
-            seg0.AddPoint(t + 65, baseLat - 0.0014, baseLon + 0.0094, 57);
-            seg0.AddPoint(t + 70, baseLat - 0.0016, baseLon + 0.0095, 55);
-            seg0.AddPoint(t + 75, baseLat - 0.0016, baseLon + 0.0095, 55); // landed
+            // Parachute descent — slow final approach (KSC terrain ~67-69m)
+            seg0.AddPoint(t + 55, baseLat - 0.0008, baseLon + 0.0091, 75);
+            seg0.AddPoint(t + 60, baseLat - 0.0012, baseLon + 0.0093, 72);
+            seg0.AddPoint(t + 65, baseLat - 0.0014, baseLon + 0.0094, 70);
+            seg0.AddPoint(t + 70, baseLat - 0.0016, baseLon + 0.0095, 69);
+            seg0.AddPoint(t + 75, baseLat - 0.0016, baseLon + 0.0095, 69); // landed
 
             // Engine events: SRB ignition and burnout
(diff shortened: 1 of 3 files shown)


vl3c (commented):
## Review: Fix engine nozzle glow after shutdown and watch button for looped segments

Overall this is a clean, well-targeted fix for both bugs. The changes are correct and consistent with the existing codebase patterns. Looks good with a few observations and one minor suggestion.

---

### Bug #18: Engine nozzle glow persists after shutdown

**EngineShutdown + ApplyHeatState (line 5215):** Correct fix. The root cause is clear: `SetEngineEmission` stops the particle FX but the `ModuleAnimateHeat` emissive material glow is a separate system that was not being reset. Adding `ApplyHeatState(state, evt, heated: false)` here is the right call. `ApplyHeatState` looks up by `evt.partPersistentId` only (not `moduleIndex`), so the EngineShutdown event's PID maps correctly to the heat info.

**Defensive cleanup on Decoupled/Destroyed (lines 5105-5107, 5115-5117):** Good defensive coding for the same-frame burnout+decouple race condition. The ordering is correct -- FX cleanup happens *before* `HidePartSubtree`/`HideGhostPart`, so the particle systems and renderers still exist when we try to stop them.

**StopEngineFxForPart / StopRcsFxForPart (lines 5505-5547):** These iterate `state.engineInfos.Values` (keyed by `ulong` encoded key) and filter by `info.partPersistentId` matching the target PID. This correctly handles multi-engine parts (multiple dictionary entries with the same PID but different encoded keys). The null guards are thorough. The particle cleanup sequence (disable emission, zero rate, Stop+Clear, disable renderers) mirrors the existing pattern in `SetEngineEmission` at power=0.

**Performance:** The linear scan over all engine/RCS infos on every decouple/destroy event is fine. Typical KSP vessels have single-digit engine counts, and these events fire once per part lifetime, not per-frame.

**Code duplication note (non-blocking):** `StopEngineFxForPart` and `StopRcsFxForPart` have identical inner loop bodies -- the only difference is which dictionary they iterate. If this pattern needs to be reused again in the future, extracting a shared helper might be worth it. But for two call sites, the explicit duplication is perfectly fine and arguably clearer.

**Completeness check -- ParachuteDestroyed:** The `ParachuteDestroyed` case (line 5140) also calls `HideGhostPart` but does *not* get the new defensive engine/RCS FX cleanup. In practice this is fine -- parachute parts don't have engines or RCS modules -- so this is not a concern. Mentioning it only for completeness.

---

### Bug #19: Watch button for looped segments

**Pause window body name initialization (lines 5593-5597):** Correct fix. The `if (string.IsNullOrEmpty(...))` guard ensures this only runs once after a cycle-change respawn (when the state is freshly constructed with null body name), and does not override values set by `SetInterpolated` in subsequent normal-playback frames. Using the last trajectory point's body name is the right data source since the ghost is positioned there.

**Removal of redundant null check (line 5598):** The old code had `if (state != null && state.reentryFxInfo != null)`. Since line 4584 already guards `if (state == null || state.ghost == null) return;`, the `state != null` check was redundant. Good cleanup.

**lastInterpolatedVelocity not initialized:** The pause window path intentionally does *not* set `lastInterpolatedVelocity` in the new body-name initialization block (only in the reentry FX block where it's set to zero). This is correct -- velocity defaults to `Vector3.zero` for a new struct, and `IsGhostOnSameBody` only uses body name, not velocity.

---

### docs/known-bugs.md

Root cause and fix descriptions are clear and accurate. Good to have these documented.

LGTM -- no issues found.

PR #39, opened 2026-03-14
Recording quality: final position capture, pad trimming, merge dialog controls

vl3c:
### Final position capture
- Force-sample a trajectory point in `OnVesselWillDestroy` so the recording ends at the actual impact point, not meters above it
- Added `RecordingVesselId` guard and InvariantCulture log formatting

### Pad/runway trimming
- `FindFirstMovingPoint` in `TrajectoryMath` trims leading stationary points at commit time (altitude change <1m AND speed <5 m/s)
- Prevents ghost from overlapping the real vessel on the pad during playback
- Orbit segments before the trim point are removed; part events are retimed to the new start (preserves shroud jettison, engine ignition, etc.)

### Short destroyed recording discard
- Destroyed recordings shorter than 10 seconds (after trimming) are auto-discarded in both the coroutine and split paths
- Prevents useless pad-failure recordings from cluttering the timeline

### Auto-merge setting
- New `autoMerge` toggle in settings (default: on)
- When on: recordings auto-commit as before
- When off: confirmation dialog appears for every recording, including crashes

### Cross-scene merge dialog
- When `autoMerge` is off, pending recordings follow the player across scenes (KSC, Tracking Station, Editor) instead of being auto-committed
- `ParsekScenario` shows a deferred merge dialog via coroutine after ~1s in any non-Flight scene
- EVA child recordings with matching parents are auto-committed silently
- `PopupDialog.DismissPopup` guard prevents double-dialogs on rapid scene changes
- Defense-in-depth safety net in `OnSave` for edge cases
- Tree snapshots preserved on scene exit when `autoMerge` is off

## Test plan
- [x] **Final position:** Crash a rocket, verify ghost ends at impact point
- [x] **Pad trimming:** Wait on pad before launching, verify ghost starts when vessel moves
- [x] **Part events:** Verify engine shrouds are off on the ghost after trimming
- [x] **Pad failure:** Rocket falls apart on pad — "Recording discarded" message, no commit
- [x] **Normal crash (>10s):** Recording auto-commits (autoMerge on) or dialog appears (autoMerge off)
- [x] **autoMerge OFF — Flight:** Dialog appears after revert
- [x] **autoMerge OFF — KSC:** Record flight, go to KSC — dialog appears
- [x] **autoMerge OFF — TS:** Record flight, go to Tracking Station — dialog appears
- [x] **autoMerge OFF — scene hop:** Ignore dialog in KSC, go to TS — dialog reappears
- [x] **autoMerge ON:** All existing behavior unchanged

### DIFF
--- a/Source/Parsek.Tests/AdaptiveSamplingTests.cs
+++ b/Source/Parsek.Tests/AdaptiveSamplingTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using UnityEngine;
 using Xunit;
 
@@ -258,6 +259,61 @@ namespace Parsek.Tests
 
             // Negative interval: 99-100 = -1, not >= 3
             Assert.False(result);
+        }
+
+        // --- FindFirstMovingPoint tests ---
+
+        private static List<TrajectoryPoint> MakePoints(params (double alt, float speed)[] specs)
+        {
+            var points = new List<TrajectoryPoint>();
+            for (int i = 0; i < specs.Length; i++)
+            {
+                points.Add(new TrajectoryPoint
+                {
+                    ut = 100 + i,
+                    altitude = specs[i].alt,
+                    velocity = new Vector3(specs[i].speed, 0, 0)
+                });
+            }
+            return points;
+        }
+
+        [Fact]
+        public void FindFirstMoving_AllStationary_ReturnsZero()
+        {
+            var points = MakePoints((78, 0.3f), (78, 0.5f), (78, 0.2f));
+            Assert.Equal(0, TrajectoryMath.FindFirstMovingPoint(points));
+        }
+
+        [Fact]
+        public void FindFirstMoving_MovingFromStart_ReturnsZero()
+        {
+            var points = MakePoints((78, 10f), (80, 15f));
+            Assert.Equal(0, TrajectoryMath.FindFirstMovingPoint(points));
+        }
+
+        [Fact]
+        public void FindFirstMoving_AltitudeChange_ReturnsCorrectIndex()
+        {
+            // Pad at alt 78, then lifts off at index 3
+            var points = MakePoints((78, 0.3f), (78, 0.5f), (78.2, 0.4f), (79.5, 2f), (85, 10f));
+            Assert.Equal(3, TrajectoryMath.FindFirstMovingPoint(points));
+        }
+
+        [Fact]
+        public void FindFirstMoving_SpeedThreshold_ReturnsCorrectIndex()
+        {
+            // Runway: same altitude, picks up speed at index 2
+            var points = MakePoints((10, 0.3f), (10, 2f), (10, 6f), (10, 15f));
+            Assert.Equal(2, TrajectoryMath.FindFirstMovingPoint(points));
+        }
+
+        [Fact]
+        public void FindFirstMoving_NullOrShort_ReturnsZero()
+        {
+            Assert.Equal(0, TrajectoryMath.FindFirstMovingPoint(null));
+            Assert.Equal(0, TrajectoryMath.FindFirstMovingPoint(new List<TrajectoryPoint>()));
+            Assert.Equal(0, TrajectoryMath.FindFirstMovingPoint(MakePoints((78, 0.1f))));
         }
     }
 }
--- a/Source/Parsek.Tests/RecordingStoreTests.cs
+++ b/Source/Parsek.Tests/RecordingStoreTests.cs
@@ -978,5 +978,95 @@ namespace Parsek.Tests
 
             Assert.Equal(4, rec.RecordingFormatVersion);
         }
+
+        // --- Stationary point trimming tests ---
+
+        private List<TrajectoryPoint> MakePadLaunchPoints()
+        {
+            // 5 stationary points on pad (alt=78, speed=0.3), then 5 moving points
+            var points = new List<TrajectoryPoint>();
+            for (int i = 0; i < 5; i++)
+                points.Add(new TrajectoryPoint
+                {
+                    ut = 100 + i, altitude = 78, velocity = new Vector3(0.3f, 0, 0), bodyName = "Kerbin"
+                });
+            for (int i = 0; i < 5; i++)
+                points.Add(new TrajectoryPoint
+                {
+                    ut = 105 + i, altitude = 80 + i * 5, velocity = new Vector3(10 + i * 5, 0, 0), bodyName = "Kerbin"
+                });
+            return points;
+        }
+
+        [Fact]
+        public void StashPending_TrimsLeadingStationaryPoints()
+        {
+            var points = MakePadLaunchPoints();
+            RecordingStore.StashPending(points, "TrimTest");
+
+            Assert.True(RecordingStore.HasPending);
+            // First 5 stationary points should be trimmed
+            Assert.Equal(5, RecordingStore.Pending.Points.Count);
+            Assert.Equal(105.0, RecordingStore.Pending.StartUT);
+        }
+
+        [Fact]
+        public void StashPending_RetimesPartEventsFromTrimmedWindow()
+        {
+            var points = MakePadLaunchPoints();
+            var partEvents = new List<PartEvent>
+            {
+                new PartEvent { ut = 100, eventType = PartEventType.ShroudJettisoned, partPersistentId = 1 },
+                new PartEvent { ut = 101, eventType = PartEventType.EngineIgnited, partPersistentId = 2 },
+                new PartEvent { ut = 107, eventType = PartEventType.Decoupled, partPersistentId = 3 }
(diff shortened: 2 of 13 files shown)


vl3c (commented):
## Code Review

Looks good. The change is correct and well-placed. A few observations:

**Correctness: safe.** `onVesselWillDestroy` fires while the `Vessel` object is still fully valid (rigidbody, orbit, position data all intact). `SamplePosition` already handles the `v.packed` branch (uses `v.obt_velocity` instead of `v.rb_velocityD`), and has a null-vessel guard at the top. No risk here.

**Ordering: correct.** The sample happens after orbit segment finalization (so `isOnRails` is cleared and the segment is closed) but before `VesselDestroyedDuringRecording = true` and `RefreshBackupSnapshot`. This is the right sequence -- the trajectory point captures position, then the flag and snapshot capture vessel state. If the order were reversed and `RefreshBackupSnapshot` somehow triggered a code path that checked `VesselDestroyedDuringRecording`, the sample would still be recorded. Good.

**On-rails destruction:** If the vessel is destroyed while on rails (e.g., lithobraking from time warp), the orbit segment gets finalized first (lines 3908-3912), then `SamplePosition` is called with `v.packed` likely still true, so it reads `v.obt_velocity`. The position (`v.latitude`/`v.longitude`/`v.altitude`) should still reflect the impact point since this is "will destroy" not "did destroy." This is fine.

**Duplicate point risk:** If the Harmony physics-frame patch happened to record a point at the exact same UT in the same frame, `SamplePosition` would add a second point at that UT. This is a pre-existing pattern -- `OnVesselGoOnRails` and `OnVesselGoOffRails` do the same thing with `SamplePosition`. The interpolation logic in `FindWaypointIndex` handles duplicate-UT points gracefully (the second point simply overrides the first during interpolation). Not a concern.

**Logging: good improvement.** The enhanced log message with lat/lon/alt is much more useful for debugging than the old generic message. Format specifiers `F4`/`F1` are reasonable precision. One nitpick: these use default string interpolation which respects system locale (per your own MEMORY.md note about `InvariantCulture`). Since this is a log message (not serialized data), locale-dependent formatting is cosmetically annoying but not a correctness bug -- a user with a comma-decimal locale would see `lat=12,3456` instead of `lat=12.3456` in the log. Worth fixing if you care about log consistency, but not blocking.

**Pre-existing: no `RecordingVesselId` guard.** The other event handlers (`OnVesselGoOnRails`, `OnVesselGoOffRails`, `OnVesselSOIChanged`) all check `v.persistentId != RecordingVesselId` after the active-vessel check. `OnVesselWillDestroy` does not. This is a pre-existing gap, not introduced by this PR. In practice, `v == FlightGlobals.ActiveVessel` should be sufficient since you only record the active vessel, but the inconsistency is worth noting for a future cleanup.

**Tests:** The `SamplePosition` method is an instance method that touches `Recording` (a list field) and KSP globals (`Planetarium`, `Krakensbane`), so it's not trivially unit-testable without mocking. The existing test coverage for the adaptive sampling path (`ShouldRecordPoint`) is solid. For this change, the in-game manual test (crash into terrain, verify ghost endpoint) described in the PR is the right approach.

**Verdict: LGTM.** No issues found. The change is minimal, correctly placed, and solves a real visual artifact (ghost trajectory ending meters above the impact point).

vl3c (commented):
## Code Review — PR #39: Recording quality: final position capture, pad trimming, merge dialog controls

Overall this is a solid and well-architected PR. The individual features are cleanly separated, the deferred dialog coroutine handles scene transitions well, and the Harmony patch on FlightResultsDialog is a smart approach. However, there is one clear issue that needs fixing before merge, plus a minor consistency item.

---

### BUG: Triple/double `ReplayFlightResultsIfPending()` calls in dialog callbacks

In `MergeDialog.cs`, all 4 button callbacks in `ShowStandaloneDialog` call `ReplayFlightResultsIfPending()` **three times** each, at different indentation levels:

```csharp
ClearPendingFlag();
            ReplayFlightResultsIfPending();   // indent 20
                ReplayFlightResultsIfPending();   // indent 24
                    ReplayFlightResultsIfPending();   // indent 28
```

This appears to be an accidental merge/rebase artifact (the varying indentation is the tell). The first call sets `Bypass = true` and calls `FlightResultsDialog.Display(msg)`. The second and third calls are no-ops because `PendingOutcomeMsg` was cleared to null by the first call. So this is **not a runtime bug** — it's functionally harmless because `ReplayFlightResults` has the `string.IsNullOrEmpty(PendingOutcomeMsg)` guard. But it is clearly unintentional dead code and should be cleaned up.

Similarly, the chain dialog callbacks in `ShowChainDialog` have **double** calls to `ReplayFlightResultsIfPending()`:

```csharp
ClearPendingFlag();
            ReplayFlightResultsIfPending();   // indent 20
                ReplayFlightResultsIfPending();   // indent 24
```

Same situation — harmless due to the null guard, but should be one call each.

The tree dialog callbacks (`ShowTreeDialog`) correctly have exactly one call each. **Fix:** reduce to a single call per callback in both `ShowStandaloneDialog` (4 callbacks x 3 calls -> 4 x 1) and `ShowChainDialog` (4 callbacks x 2 calls -> 4 x 1).

---

### Minor: `autoMerge` default vs Defaults button mismatch

`ParsekSettings.autoMerge` field default is `false` (line 34 of ParsekSettings.cs), but the "Defaults" button in `ParsekUI.DrawSettingsWindow` sets `s.autoMerge = true` (line 1878 of ParsekUI.cs). This means a new game starts with autoMerge=off (dialog appears), but clicking "Defaults" switches it to on (auto-commit). These should be consistent. Given the PR description says "default: off", the Defaults button should probably set it to `false`.

---

### IsAutoMerge cache: safe

The cached `cachedAutoMerge` approach is sound. The only scenario where the cache could become stale is: player changes the setting in KSP's Game Settings menu, then a scene transition happens where `ParsekSettings.Current` is temporarily null. But `IsAutoMerge` updates the cache on every successful read, so by the time the player actually produces a recording and triggers a commit path, the cache will have been refreshed many times. No issue here.

---

### Deferred dialog coroutine and scene changes: well handled

The 60-frame wait in `ShowDeferredMergeDialog` is reasonable. The `mergeDialogPending` flag is reset at the top of `OnLoad`, and `OnLoad` also removes the input lock — so if the player changes scenes during the wait, the coroutine will be destroyed with the old `ParsekScenario` instance (Unity coroutines are tied to MonoBehaviour lifecycle), and the new `OnLoad` will start a fresh coroutine if needed. The `!mergeDialogPending` guard prevents double-starting the coroutine, and `PopupDialog.DismissPopup("ParsekMerge")` prevents stacking dialogs. Well handled.

---

### Input lock management: robust

The lock is set in `LockInput()` before dialog spawn, cleared in `ClearPendingFlag()` from every button callback, and cleared as a safety net in `OnLoad`. The only theoretical stuck-lock path is if `PopupDialog.SpawnPopupDialog` throws after `LockInput()`, but KSP's PopupDialog doesn't throw. The `OnLoad` safety net covers the case where a scene change destroys the dialog without a button click. Solid.

---

### Harmony patch — Bypass flag: safe

The `Bypass` flag in `FlightResultsPatch` is set to `true` before calling `FlightResultsDialog.Display(msg)`, and the prefix clears it immediately on entry. Since Unity is single-threaded for game logic, there's no race condition. If `Display` somehow fails/throws, the flag stays true, but the next genuine `Display` call would just pass through (benign — KSP shows its dialog, which is the default behavior). Safe.

---

### Trimming + part event retiming: correct

Events from the trimmed window get their `ut` set to `trimUT`, firing at the very beginning of ghost playback. This is the right behavior — a shroud jettison that happened on the pad should show as "jettisoned" from frame 1 of the ghost. The orbit segment removal (`endUT <= trimUT`) correctly preserves segments that span the trim boundary. A segment with `startUT` before the recording's `StartUT` could remain, but orbit playback only uses segments at or after the current UT, so it's harmless.

One subtlety: `FindFirstMovingPoint` returns 0 for an all-stationary recording, meaning no trimming occurs and the full recording is kept. This is correct behavior for a deliberate stationary recording.

---

### Minor: Locale in log messages

Several log messages use `$"{val:F1}"` interpolation (e.g., `flightDuration:F1`, `maxDist:F0`, `trimUT:F1`) which respects system locale. Per project conventions, `InvariantCulture` is preferred. These are log messages only (not serialized data), so it's cosmetic. Not blocking.

---

### Verdict

Two items to fix before merge:
1. **Duplicate `ReplayFlightResultsIfPending` calls** — reduce to one per callback in `ShowStandaloneDialog` and `ShowChainDialog`
2. **Defaults button** — `s.autoMerge = true` should be `s.autoMerge = false` to match the field default

Everything else (Harmony patch, coroutine lifecycle, input locking, trimming/retiming, IsAutoMerge caching) is well-designed and correct.

PR #40, opened 2026-03-14
Fix save contamination: recordings leaking across save games

vl3c:
## Summary

- **CommittedTrees not cleared on save switch**: `committedTrees.Clear()` was inside `if (treeNodes.Length > 0)` in OnLoad's initial-load path. When switching from save A (has tree recordings) to save B (no trees), save A's trees persisted in the static list and got serialized to save B's .sfs on next OnSave. Moved `Clear()` before the conditional.
- **Pending recordings/trees leak across saves**: Static `pendingRecording`/`pendingTree` survived save switches. Auto-commit logic at end of OnLoad would commit save A's pending data into save B. Added cleanup at start of initial-load path.
- **Rewind statics leak across saves**: `IsRewinding` could survive a save switch, causing the rewind branch to execute instead of normal initial load. Added stale rewind flag cleanup.
- **Diagnostic warning**: OnSave now logs a warning if `HighLogic.SaveFolder` doesn't match the scenario's loaded save folder (log-only, no behavioral change).
- 9 new unit tests in `SaveContaminationTests.cs`.

## Test plan

- [x] `dotnet build` from `Source/Parsek/` succeeds
- [x] `dotnet test` from `Source/Parsek.Tests/` — all new tests pass, no regressions
- [x] In-game: load save A with tree recordings, switch to save B (no trees), verify save B's .sfs has no save A recordings
- [x] In-game: start recording in save A, switch to save B before committing, verify pending recording is discarded (not auto-committed to save B)

### DIFF
--- a/Source/Parsek.Tests/SaveContaminationTests.cs
+++ b/Source/Parsek.Tests/SaveContaminationTests.cs
@@ -1,0 +1,155 @@
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Parsek.Tests
+{
+    [Collection("Sequential")]
+    public class SaveContaminationTests : IDisposable
+    {
+        private readonly List<string> logLines = new List<string>();
+
+        public SaveContaminationTests()
+        {
+            RecordingStore.SuppressLogging = true;
+            RecordingStore.ResetForTesting();
+            MilestoneStore.SuppressLogging = true;
+            MilestoneStore.ResetForTesting();
+            GameStateStore.SuppressLogging = true;
+            ParsekLog.ResetTestOverrides();
+            ParsekLog.SuppressLogging = false;
+            ParsekLog.TestSinkForTesting = line => logLines.Add(line);
+        }
+
+        public void Dispose()
+        {
+            ParsekLog.ResetTestOverrides();
+            ParsekLog.SuppressLogging = true;
+            RecordingStore.SuppressLogging = true;
+            RecordingStore.ResetForTesting();
+            MilestoneStore.ResetForTesting();
+        }
+
+        // --- IsSaveFolderMismatch tests ---
+
+        [Fact]
+        public void IsSaveFolderMismatch_DifferentFolders_ReturnsTrue()
+        {
+            Assert.True(ParsekScenario.IsSaveFolderMismatch("saveA", "saveB"));
+        }
+
+        [Fact]
+        public void IsSaveFolderMismatch_SameFolders_ReturnsFalse()
+        {
+            Assert.False(ParsekScenario.IsSaveFolderMismatch("saveA", "saveA"));
+        }
+
+        [Fact]
+        public void IsSaveFolderMismatch_NullScenarioFolder_ReturnsFalse()
+        {
+            Assert.False(ParsekScenario.IsSaveFolderMismatch(null, "saveB"));
+        }
+
+        [Fact]
+        public void IsSaveFolderMismatch_EmptyScenarioFolder_ReturnsFalse()
+        {
+            Assert.False(ParsekScenario.IsSaveFolderMismatch("", "saveB"));
+        }
+
+        [Fact]
+        public void IsSaveFolderMismatch_BothNull_ReturnsFalse()
+        {
+            Assert.False(ParsekScenario.IsSaveFolderMismatch(null, null));
+        }
+
+        // --- CommittedTrees contamination scenario ---
+
+        [Fact]
+        public void CommittedTrees_SurvivesRecordingsClear_DemonstratesBugScenario()
+        {
+            // Simulate save A: commit a tree (adds to both CommittedTrees and CommittedRecordings)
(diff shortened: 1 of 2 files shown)


vl3c (commented):
## Code Review: Fix save contamination

Reviewed the full diff plus surrounding context in `ParsekScenario.cs` (OnLoad/OnSave flow, revert detection, auto-commit path) and `RecordingStore.cs` (pending/committed state, rewind flags, `ResetForTesting`).

### Verdict: Looks good. Ship it.

All three bugs are real, the fixes are correct, and the placement in the OnLoad flow is right. Details below.

---

### 1. Correctness

**CommittedTrees fix** -- Moving `committedTrees.Clear()` out of the `if (treeNodes.Length > 0)` block is the right fix. The old code silently leaked trees from save A into save B whenever save B had zero trees. The new code unconditionally clears before loading, which matches the existing pattern for `recordings.Clear()` on the line above.

**Pending state cleanup** -- Correct placement: runs inside the `!initialLoadDone` (initial-load) path, *after* `initialLoadDone = true` is set but *before* `recordings.Clear()` and the deserialization loop. Since pending state can only exist from a *previous* save session (the auto-commit path on line 500-517 fires for non-revert scene changes *within* the same save), the only way to reach initial-load with pending state is a cross-save switch. Discarding is the right call.

**Rewind flag cleanup** -- Correct. If `IsRewinding` leaked from save A into save B's initial load, the rewind branch (line 282) would execute and `return` early, skipping the entire deserialization. The cleanup prevents this.

### 2. Safety

**No data loss risk.** The cleanup only fires on the initial-load path (save switch), not on reverts or scene changes. Within a single save session, `initialLoadDone` is `true`, so the revert/scene-change path handles pending state via auto-commit (line 500-517) or merge dialog -- unaffected by this change.

**`DiscardPending()` deletes sidecar files** (`DeleteRecordingFiles`), which is correct -- those files belong to the *old* save's recording directory. If the pending recording had already been written to save A's Parsek/Recordings/ folder, the files get cleaned up. If not, `DeleteRecordingFiles` is a no-op. Either way, safe.

**OnSave diagnostic** is log-only, no behavioral change. `IsSaveFolderMismatch` correctly returns `false` when `scenarioSaveFolder` is null (before first OnLoad), avoiding false positives.

### 3. Edge cases

**First load (scenarioSaveFolder is null):** `IsSaveFolderMismatch(null, x)` returns `false` -- correct, no false warning.

**Save switch during active recording:** `FlightRecorder` holds its own state, but if the user somehow switches saves mid-recording (e.g., Alt+F4 and reload different save), the pending state from the aborted recording gets cleaned up here. Good.

**`IsRewinding` cleanup is incomplete (minor):** The PR clears `IsRewinding`, `RewindUT`, and `RewindAdjustedUT`, but not `RewindReserved`, `RewindBaselineFunds`, `RewindBaselineScience`, or `RewindBaselineRep`. These are harmless leftovers since the rewind coroutine only runs when `IsRewinding` was true at the time the rewind OnLoad path executed (line 282), and that path won't execute after the fix. But for symmetry with the existing cleanup in the rewind-success path (lines 323-327) and the error paths in `InitiateRewind` (lines 1154-1157, 1185-1188), consider also zeroing these. Not a bug, just defense-in-depth.

### 4. Code quality

- Logging is good: `ParsekLog.Warn` for the cross-save pending/tree/rewind discards, `ParsekLog.Info` for the tree clear. Follows the project's "if it didn't get logged, it didn't happen" convention.
- The `IsSaveFolderMismatch` extraction as `internal static` is clean -- testable without Unity, consistent with other extracted methods (`CheckParachuteTransition`, `IsSaveFolderMismatch`, etc.).
- Comments explain *why* the clears are needed, not just *what* they do.

### 5. Test coverage

The 9 tests cover the core scenarios well:
- `IsSaveFolderMismatch` -- all edge cases (null, empty, same, different, both-null)
- `CommittedTrees_SurvivesRecordingsClear` -- demonstrates the bug existed
- `CommittedTrees_ExplicitClear` -- demonstrates the fix works
- `DiscardPendingTree` / `DiscardPending` -- verifies cleanup

**Missing test scenarios** (nice-to-have, not blocking):
- No test that asserts `IsRewinding` cleanup. Could add: set `IsRewinding = true` + `RewindUT = 123`, call the cleanup logic, assert both are reset.
- No integration-level test simulating the full OnLoad sequence (load save A with trees, then load save B without trees, verify CommittedTrees is empty). This would require more test infrastructure and is better covered by the manual in-game test plan.
- Log assertion tests: the pending/tree discard paths emit `ParsekLog.Warn` -- could assert those appear in `logLines`. Follows the project's log-assertion pattern from `RewindLoggingTests.cs`.

### Summary

The fix is correct, safe, and well-tested. The three contamination vectors (CommittedTrees, pending recordings/trees, rewind flags) are all addressed at the right point in the load lifecycle. The only minor suggestion is zeroing the remaining rewind baseline fields for symmetry, but it's not a functional issue.

PR #159, opened 2026-04-09
Fix 3 playtest-failing in-game tests (#257, #258, #259)

vl3c:
## Summary

Fixes three in-game test failures from the latest playtest:

- **#257 `OrbitSegmentBodiesValid`** — test assertion bug. Hyperbolic escape orbits (eccentricity > 1) legitimately have negative semi-major axis. Test now skips the `SMA > 0` check when `e > 1`.
- **#258 `CommittedRecordingsHaveValidData`** — code bug. A quickload during recording of Bob Kerman's EVA reset game time by ~33s, but the recorder kept appending points without trimming the pre-quickload data. Added `TrimRecordingToUT` to `FlightRecorder`; both `CommitRecordedPoint` and `SamplePosition` now detect time regression (>1s backward) and trim stale points, part events, and orbit segments.
- **#259 `OrbitalRecordingsHaveTerminalOrbit`** — code bug. Three paths (`CaptureSceneExitState`, `ChainSegmentManager` vessel-switch, `BackgroundRecorder` debris end) set `TerminalStateValue` to Orbiting/Docked without calling `CaptureTerminalOrbit`. `FinalizeIndividualRecording` then skipped orbit capture because the terminal state was already set. Added `CaptureTerminalOrbit` calls at all three source paths and a defensive backfill in `FinalizeIndividualRecording` for orbital/suborbital/docked recordings with missing `TerminalOrbitBody`.

Also includes two earlier docs commits adding/updating investigation items #256-#262.

## Test plan

- [x] 5 new `TrimRecordingToUT` unit tests
- [x] Full unit test suite: 5202 pass
- [x] Clean build
- [ ] Re-run in-game playtest to confirm #257 and #259 pass
- [ ] #258 is prospective — existing corrupted recording `ab105395ae5547b0b70c1eb9bb41ca9f` still fails until deleted or re-recorded

## Notes

- `CaptureTerminalOrbit` visibility changed from `private static` to `internal static` so `BackgroundRecorder` and `ChainSegmentManager` can call it directly.
- The branch was originally based on `maintenance`, which has since been merged and deleted from the remote. Rebased onto `main`.


vl3c:
## Review follow-ups (commit `77bce7c`)

Addressed the actionable findings from the review:

**Correctness fixes**
- `TrimRecordingToUT` comment now matches the implementation (linear, not binary)
- Extracted `FlightRecorder.TimeRegressionThresholdSeconds = 1.0` const, shared between `CommitRecordedPoint` and `SamplePosition`
- Note added about `lastRecordedUT` staleness inside the trim
- **#259 backfill order-of-operations**: previously only fell back to `PopulateTerminalOrbitFromLastSegment` when `v == null`. Now checks `TerminalOrbitBody` AFTER the live capture attempt and falls back when still empty (covers the case where `v != null` but `CaptureTerminalOrbit` silently declines due to orbit == null or situation outside the accepted set). Logs the capture source for diagnostics.
- `BackgroundRecorder` empty-branch now logs a warn ("tree recording not found — upstream lifecycle bug") instead of silently skipping

**Project conventions**
- `TrimRecordingToUT` warn log now uses `InvariantCulture` for all double formatting (was the project's documented locale-safety rule). Breaks comma-locale log parsing otherwise.
- Settings file moved from `GameData/Parsek/settings.cfg` to `GameData/Parsek/PluginData/settings.cfg` per KSP convention — excluded from ModuleManager's patch cache, standard location for runtime-written mod state. `FileIOUtils.SafeWriteConfigNode` already creates parent directories.

**Tests added** (`ReviewFollowupTests.cs`, 6 new, 5213 total):
- `TrimRecordingToUT_UsesInvariantCultureForLogging` — runs under `de-DE` current culture, verifies period decimals in the log
- `TimeRegressionThreshold_IsOneSecond` — pins the constant value
- `PopulateTerminalOrbitFromLastSegment_EmptyBody_FallsBack` / `NullOrbitSegments_LeavesEmpty` — anchors the scenarios referenced by the new backfill order-of-operations
- `SeedEventSkipPredicate_*` — documents the `PartEvents.Count > 0` predicate used by `BackgroundRecorder.InitializeLoadedState`

**Not added (deferred to in-game tests)**
- `PauseAllAudio` / `UnpauseAllAudio` null-guard coverage. The methods' IL references `UnityEngine.AudioSource` via the `audioInfos` foreach, which trips xUnit's *"ECall methods must be packaged into a system module"* even for a null-state early return. Can't exercise from xUnit without a Unity mocking layer.
- `FinalizeIndividualRecording` full backfill path (needs a live `ParsekFlight` + `Vessel`) — covered by the in-game test verification scheduled for the playtest round.

**Not addressed (flagged as process, not blocking)**
- **Scope expansion beyond the 3 original test fixes** (bug A seed-event skip, audio pause, settings persistence, design doc). Acknowledged — the expansion happened incrementally during investigation as related issues surfaced, and each was approved before commit. Splitting now would require a disruptive rebase for questionable benefit. Future bundles will be smaller.
- **Time warp sub-second jitter false-positive on the 1s threshold.** Unlikely per reviewer ("worth noting in a comment"). I added a comment on the constant explaining the choice; worth revisiting if playtest logs show spurious trims.
- **Parabolic `e == 1.0` edge case** on the `OrbitSegmentBodiesValid` test. Measure zero in practice — reviewer's "probably fine" assessment stands.
- **Existing corrupted recording `ab105395ae5547b0b70c1eb9bb41ca9f`** will still fail the in-game test until manually deleted. Fix is prospective-only per the original PR description.

Clean build, 5213/5213 unit tests pass.

vl3c:
## Round-2 review follow-ups (commit `a99be6a`)

All four observations addressed:

**1. Tautological tests deleted** — `SeedEventSkipPredicate_RecordingHasPartEvents_SkipsSeed` and `_EmptyRecording_EmitsSeed` re-implemented the branch condition in the test itself (zero regression value). Removed.

**2. InvariantCulture log test tightened** — now asserts the exact substring `"from -1.0 to 105.5"` (pins both `lastRecordedUT` and `newUT` formatting). `lastRecordedUT` defaults to `-1` per `FlightRecorder.cs:193`, so the expected sentinel is explicit. If the initial value ever changes, the test will force us to notice.

**3. Backfill "both sources declined" log** — split the branch into three paths:
```csharp
if (!string.IsNullOrEmpty(rec.TerminalOrbitBody))         // live vessel win
    ParsekLog.Info(...from live vessel...);
else {
    PopulateTerminalOrbitFromLastSegment(rec);
    if (!string.IsNullOrEmpty(rec.TerminalOrbitBody))     // segment fallback win
        ParsekLog.Info(...via orbit-segment fallback...);
    else                                                  // both declined
        ParsekLog.Warn(...backfill declined...orbitSegments=0...);
}
```
No more trailing empty `TerminalOrbitBody=` in the log. Warn level on the both-declined path surfaces the degraded state loudly in playtest logs.

**4. Tracked follow-up #265** — added to `docs/dev/todo-and-known-bugs.md` with the full context on the xUnit/Unity audio coverage gap. Lists the four specific code paths blocked by the "ECall methods must be packaged" error (audio pause/unpause, engine loop, background seeder predicate, `FinalizeIndividualRecording` backfill) and proposes `InGameTest` categories as the fix.

5211/5211 tests pass (dropped 2 tautological + kept 4 meaningful review-followup tests).

PR #160, opened 2026-04-09
Quickload-resume recording (Bug C / H / I)

vl3c:
## Summary

Fixes the root cause of the quicksave+quickload breakage observed in the 2026-04-09 KerbalX playtest (Bug C), plus its two cascading effects:

- **Bug C** — quicksave + quickload during an active recording fragments the mission: tree commits prematurely, ~2 minutes of untracked flight until the user manually restarts recording, crew scattered across unlinked standalone recordings.
- **Bug H** — ghost orbits are partial arcs (vessel disappears mid-flight) because the committed tree only has trajectory up to the quicksave UT.
- **Bug I** — wrong crew spawns at mission end because the original crew was captured into orphan recordings.

Both H and I resolve automatically when C is fixed.

## Design

Full design doc added at `docs/dev/plans/quickload-resume-recording.md` — design was reviewed by a clean Opus agent before implementation, and the implementation was reviewed by another clean Opus agent after.

### Three compounding defects

- **C1** — `ParsekFlight.FinalizeTreeOnSceneChange` called `CommitTreeRevert` on every FLIGHT→FLIGHT reload (including quickloads), finalizing the active tree **before** OnLoad could distinguish revert from quickload.
- **C2** — `ParsekScenario.OnSave` only persisted committed recordings; the active tree was never serialized, so even if we refused to commit, the scene reload would still lose the data.
- **C3** — `ParsekScenario.OnLoad`'s `isRevert` condition conflated FLIGHT→FLIGHT with revert via `|| isFlightToFlight`, making quickloads look like reverts.

### Fix (atomic)

All three plus the new machinery land together (see design doc's "Why atomic" section — intermediate states are all broken).

- **New `PendingTreeState` enum** `{Finalized, Limbo}` on the existing `pendingTree` slot (reviewer pushed back on a separate `limboTree` slot — consolidated into a state flag instead).
- **`StashActiveTreeAsPendingLimbo`** replaces `CommitTreeRevert` on the FLIGHT→FLIGHT path. Flushes recorder buffers into the tree, aborts any `pendingSplitRecorder`, stashes as Limbo without finalizing terminal state.
- **`SaveActiveTreeIfAny`** writes the active tree as an additional `RECORDING_TREE` node flagged `isActive=True`, guarded on `FLIGHT && activeTree && recorder`.
- **`TryRestoreActiveTreeNode`** in OnLoad (after settings apply, before revert detection) parses the active-tree node back into pending-Limbo.
- **`isRevert` cleanup**: removed `|| isFlightToFlight` clause.
- **Dispatch**: Limbo + `isRevert` → `FinalizePendingLimboTreeForRevert` + auto-commit path; Limbo + `!isRevert` → schedule `RestoreActiveTreeFromPending` coroutine for OnFlightReady.
- **`RestoreActiveTreeFromPending`** coroutine: 3-second vessel name-match, EVA-boarded parent fallback (walks `ParentRecordingId` up to 5 levels), PID remap, `PopPendingTree` (non-destructive), `StartRecording(isPromotion: true)` — relies on the Bug A fix already in `fix/playtest-failures` for seed-event skip.
- **Cold-start path**: TryRestoreActiveTreeNode also runs in the `initialLoadDone=false` branch so a fresh KSP launch loading a save with an in-flight tree resumes correctly.

## Commits

- `df78255` Design doc (reviewed by clean Opus agent; review feedback merged into doc before implementation started)
- `d1705bd` Atomic Step 1: core plumbing (5 files, 662+ insertions)
- `82a0c58` Post-implementation review fixes: cold-start restore, EVA-boarded fallback, defensive active-tree skip

## Test plan

- [x] 26 new `QuickloadResumeTests` unit tests covering `PendingTreeState` transitions, `PopPendingTree`, `IsActiveTreeNode`, `TryRestoreActiveTreeNode` dispatch, resume-hint parsing, and `isRevert` logic after the clause removal
- [x] Full xUnit suite: 5233 pass (5207 baseline + 26 new)
- [x] Clean build (0 warnings)
- [ ] **In-game playtest** — the primary validation. Need to run through the 2026-04-09 KerbalX scenario end-to-end:
  - Launch KerbalX, quicksave during ascent, quickload → recording should resume appending to the same chain segment (same `recId`, points at last kept UT ≤ quicksave UT)
  - Revert to Launch during recording → tree finalized normally, merge dialog appears if autoMerge off
  - Vessel switch during recording → tree survives, old vessel's recording moved to BackgroundMap
  - Quickload mid-chain-continuation (undock window) → `pendingSplitRecorder` aborted cleanly
  - Quickload → quickload → verify idempotent
  - Quit KSP mid-flight, relaunch, Resume Saved Game → cold-start restore path
- [ ] In-game tests under `InGameTests/RuntimeTests.cs` — deferred to a follow-up commit once the above manual playtest validates the infrastructure

## Known deferred items

Flagged by the post-implementation review, deferred because they need design discussion rather than quick fixes:

1. **Snapshot preservation through revert finalization.** The old `CommitTreeRevert` preserved vessel snapshots so the merge dialog could offer spawning. `FinalizePendingLimboTreeForRevert` doesn't — by OnLoad time, vessel refs are gone, so we fall back to whatever snapshots were in the tree when OnSave wrote them. Potential regression for autoMerge=off + revert + dialog-driven respawn. Assessing after playtest.
2. **`restoring` reentrancy flag.** `onVesselChange` fires when the quickloaded scene finishes loading, potentially mutating tree state during the 3-second wait window in the restore coroutine. Needs a guard flag. Tracked as a follow-up.
3. **Multi-tree hard assertion.** `StashPendingTree(Limbo)` currently warns+overwrites when a prior pending tree exists; the design called for a harder fail. Acceptable per "fail loud" principle; can strengthen if it becomes a real problem.

## Stacked on

This PR is stacked on top of [#159](https://github.com/vl3c/Parsek/pull/159) (`fix/playtest-failures`), which contains the Group 1 fixes (A, D, F + the Bug A seed-event skip that this PR's resume path depends on). Target branch is set to `fix/playtest-failures`; after that merges, this will rebase onto `main` for the final merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,27 @@ All notable changes to Parsek are documented here.
 - **Fix ghost audio not muted when KSP pause menu (ESC) is open.** Ghost `AudioSource`s had no pause handling — engine sounds continued playing during the ESC menu. Subscribe to `GameEvents.onGamePause` / `onGameUnpause` in `ParsekFlight`, call new `GhostPlaybackEngine.PauseAllGhostAudio` / `UnpauseAllGhostAudio` which iterate all primary + overlap ghost states. New `GhostPlaybackLogic.PauseAllAudio` / `UnpauseAllAudio` use `AudioSource.Pause()` / `.UnPause()` (preserves playback position, unlike the existing `MuteAllAudio` which calls `Stop()`).
 - **Fix ghost camera cutoff not persisted across rewind or KSP restart.** `ParsekSettings` lives inside KSP's `GameParameters`, which get reset on every quickload (including `parsek_rw_*` rewinds). User-set values for `ghostCameraCutoffKm` were lost on every rewind. New `ParsekSettingsPersistence` store writes user changes to `GameData/Parsek/PluginData/settings.cfg` (KSP convention for runtime mod state — excluded from ModuleManager's patch cache) and reads them back in `ParsekScenario.OnLoad` immediately after the settings node is available. Survives rewind, save/load, scene transitions, and KSP session restart.
 
+#### Quickload-resume recording (Bug C + cascading H, I)
+
+Observed in the 2026-04-09 KerbalX playtest: quicksave + quickload during an active recording fragmented the mission across multiple unlinked recordings, produced partial ghost orbits (Bug H), and scattered crew across orphan recordings (Bug I). Both H and I are cascading effects of the same root cause — the tree was being prematurely finalized on every FLIGHT→FLIGHT scene reload.
+
+Three compounding defects:
+
+- **`ParsekFlight.FinalizeTreeOnSceneChange` called `CommitTreeRevert` on any FLIGHT→FLIGHT reload**, finalizing the active tree before `OnLoad` could distinguish revert from quickload.
+- **`ParsekScenario.OnSave` never serialized the active tree**, so even if we refused to commit in memory, the scene reload would lose it.
+- **`ParsekScenario.OnLoad`'s `isRevert` condition conflated FLIGHT→FLIGHT with revert** via `|| isFlightToFlight`.
+
+Atomic fix (landed together — the intermediate states are all broken):
+
+- New `PendingTreeState` enum `{Finalized, Limbo}` on the existing `pendingTree` slot. Stash-as-Limbo preserves the tree data without finalization; `OnLoad` decides between restore-and-resume (quickload) and finalize-and-commit (revert or vessel switch).
+- `OnSave` writes the active tree as a `RECORDING_TREE` node flagged `isActive=True`, gated on `LoadedScene == FLIGHT && activeTree && recorder`. Recorder buffers are flushed into the tree before serialization. Resume hint (rewind save filename) serialized alongside.
+- `OnLoad` runs new `TryRestoreActiveTreeNode` after `ParsekSettingsPersistence.ApplyTo` but before revert detection; stashes the loaded active tree into pending-Limbo. Dedupes against `committedTrees` to handle the in-flight → TS commit → quickload edge case. Runs on both `initialLoadDone=true` (in-session quickload) and `initialLoadDone=false` (cold-start "Resume Saved Game") branches.
+- Removed `|| isFlightToFlight` from the `isRevert` condition — only epoch regression or count regression count as revert indicators now.
+- New `StashActiveTreeAsPendingLimbo` replaces `CommitTreeRevert` on the FLIGHT→FLIGHT path. Flushes recorder buffers, finalizes background orbit segments, aborts any `pendingSplitRecorder` (continuation windows can't survive scene reload cleanly), then stashes as Limbo without setting terminal state.
+- New `RestoreActiveTreeFromPending` coroutine runs from `OnFlightReady`: 3-second vessel name match, EVA-boarded-parent fallback (walks `ParentRecordingId` up to 5 levels), PID remap, fresh `FlightRecorder` attached to the restored tree, `StartRecording(isPromotion: true)` relies on the Bug A seed-event skip to avoid duplicate events at the quickload UT. Non-destructive `PopPendingTree` preserves sidecar files.
+- New `FlightRecorder.LastRecordedAltitude` cache so `HandleSoiAutoSplit` can classify `fromPhase` correctly even after `FlushRecorderIntoActiveTreeForSerialization` clears the `Recording` buffer mid-flight. Updated by `CommitRecordedPoint`, `SamplePosition`, and `InsertBoundaryAnchorAndSnapshot` (for chain continuation anchors).
+- Vessel-switch scene reloads route through `FinalizePendingLimboTreeForRevert` to match mainline behavior — the new active vessel by definition can't match the tree's recorded vessel by name, so restoring would trap in the 3-second wait. Proper tree-preservation-on-switch is a follow-up (#266).
+
 ### Bug Fixes (Mun Mission Playtest)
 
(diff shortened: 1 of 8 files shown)


vl3c:
## Review follow-ups (commit `31621d6`)

All 7 review findings addressed:

**1. HIGH — buffer clearing SOI altitude read → `FlightRecorder.LastRecordedAltitude` cache**
New `internal double LastRecordedAltitude { get; private set; } = double.NaN` field, populated by `CommitRecordedPoint` and `SamplePosition` alongside `lastRecordedUT` / `lastRecordedVelocity`. `HandleSoiAutoSplit` now reads the cached field instead of peeking at `recorder.Recording[last].altitude`, so a `FlushRecorderIntoActiveTreeForSerialization` landing between physics frames and an SOI crossing no longer misclassifies `fromPhase` as `"exo"`.

**2. MEDIUM — `resumeBoundaryAnchorUT` dead write removed**
Dropped the write in `SaveActiveTreeIfAny`, the read in `TryRestoreActiveTreeNode`, and the `pendingActiveTreeResumeBoundaryAnchorUT` static field entirely. Legacy saves with the key are tolerated (ignored, not parsed). A comment on both sides documents why: reconstructing a `TrajectoryPoint` from a bare UT is impossible, and missing anchor just produces one extra boundary point on the next chain continuation.

**3. MEDIUM — vessel-switch dispatch regression**
`isVesselSwitch` now routes through `FinalizePendingLimboTreeForRevert` alongside `isRevert`, matching mainline `CommitTreeRevert` behavior. Without this, `RestoreActiveTreeFromPending` would wait 3s for a name match against the new (different) active vessel, fail, and abandon the tree in Limbo — a regression relative to mainline. Tree-preservation-on-vessel-switch is a proper improvement deferred to a follow-up; commented in-line and in the design doc's Risks section.

**4. MEDIUM — multi-tree edge case**
New `RemoveCommittedTreeById` helper called from `TryRestoreActiveTreeNode` before stashing. On the flight→quicksave→TS-commit→quickload sequence, the in-memory committedTrees entry for the same tree id is removed so the active version (from the disk save) is the single source of truth. Avoids duplicate-id collision on the next `OnSave`.

**5. LOW — overwrite warning on every quickload**
`TryRestoreActiveTreeNode` now calls `RecordingStore.PopPendingTree` (non-destructive, no warning) before stashing the fresh disk version. The expected-overwrite path (`StashActiveTreeAsPendingLimbo` → `TryRestoreActiveTreeNode`) is now silent. Test `TryRestoreActiveTreeNode_NoOverwriteWarningOnExpectedOverwrite` asserts no warn log line fires.

**6. LOW — `RewindSaveFileName` encapsulation**
Reverted `internal set` back to `private set`. Added `FlightRecorder.SetRewindSaveFileNameForRestore(fileName, reason)` method that does the assignment and logs the transition. Matches the project convention for state-mutation logging. Only caller is `RestoreActiveTreeFromPending`.

**7. LOW — `ScheduleActiveTreeRestoreOnFlightReady` bare static bool**
Acknowledged and left as-is per the design doc's Risks section — setting the flag twice is idempotent, and a proper reentrancy guard is tracked as a follow-up.

**Tests** (now 5240 total, 5233 baseline + 7 meaningful net new):
- `TryRestoreActiveTreeNode_ParsesResumeRewindSave` / `_NoResumeRewindSave_ClearsStaleValue` (replaced the old `_ParsesResumeHints` pair after dropping the BoundaryAnchor UT field)
- `TryRestoreActiveTreeNode_DoesNotWriteDeadBoundaryAnchorUT` — legacy save tolerance
- `TryRestoreActiveTreeNode_DedupeCommittedTreeById` — #4 fix
- `TryRestoreActiveTreeNode_NoOverwriteWarningOnExpectedOverwrite` — #5 fix, asserts absence of the warn log
- `LastRecordedAltitude_DefaultsToNaN` — #1 default-value anchor
- `LimboDispatch_{Revert, VesselSwitch, Quickload}_*` — mirror of the dispatch decision (pure logic), covering #3

**Not addressed in this commit (tracked follow-ups)**
- Tree-preservation-on-vessel-switch — proper fix requires background-map plumbing; deferred
- Snapshot preservation through revert finalization (from the design doc)
- Restore coroutine reentrancy flag (`restoring` guard)
- Full in-game test coverage for the flow

Clean build, 5240/5240 tests pass.

vl3c:
## Round-3 review follow-up (commits \`51422a5\`, \`b0d806c\`, \`ff1f0cd\`)

All items addressed:

**Gap: `LastRecordedAltitude` not updated on boundary anchor insert** — one-line fix in `InsertBoundaryAnchorAndSnapshot` (FlightRecorder.cs) mirrors the existing `lastRecordedVelocity = anchor.velocity` pattern. Commit \`51422a5\`.

**Deferred items tracked in the canonical TODO file** (`docs/dev/todo-and-known-bugs.md`):

- **#266** Tree-preservation on vessel switch — current code finalizes on switch (matches mainline), proper fix keeps the tree alive with the old vessel backgrounded
- **#267** Restore coroutine reentrancy guard (\`restoring\` flag)
- **#268** Snapshot preservation through revert finalization
- **#269** In-game test coverage for quickload-resume flow — the Unity-dependent gaps in the unit test suite (quickload resume, revert finalize, vessel switch, double-F9, cold-start, EVA-boarded fallback)
- **#270** Sidecar file (.prec) version staleness across save points — latent pre-existing issue flagged by reviewer, not caused by this PR but more reachable with quickload-resume

**CHANGELOG updated for 0.7.2** — new "Quickload-resume recording (Bug C + cascading H, I)" subsection under the existing "KerbalX + Butterfly Rover Playtest" section. Covers the three compounding defects, the atomic fix (PendingTreeState enum, active-tree serialization, OnLoad dispatch, restore coroutine, LastRecordedAltitude cache, vessel-switch behavior), and the scoping note on #266.

Also deduped: the first six #159 entries were temporarily in this branch's CHANGELOG but have been removed — they'll land via PR #159 and appear above the quickload-resume subsection after #159 merges and #160 rebases. Prevents duplicate-entries on rebase. Placeholder HTML comment in the file explains the expected state for future readers of the mid-stack diff.

5240/5240 tests pass. Build clean.

PR #165, opened 2026-04-09
Fix #264: EVA kerbals spawn on top of parent vessel instead of recorded endpoint

vl3c:
## Summary

Fixes #264 (in-flight spawn path). During the 2026-04-09 Butterfly Rover playtest, Valentina EVA'd from a rover and walked ~170 m away; on commit her ghost spawned on top of the rover. Root cause confirmed via decompilation; fix routes EVA (and breakup-continuous) spawns through the existing `SpawnAtPosition` path so the ORBIT subnode is rebuilt from the endpoint state vectors, adds an override for FLYING to LANDED/SPLASHED based on terminal state, removes the `!isEva` guard on collision checks, and adds a subdivided trajectory walkback with 1.5 m linear sub-steps when the recorded endpoint overlaps a loaded vessel.

## Root cause (decompile)

`ProtoVessel.Load` writes the snapshot `lat`/`lon`/`alt` to `vesselRef.transform.position` correctly on frame 0 — the transform is initially at the recorded endpoint. But `OrbitDriver.updateMode` is initialized from `Vessel.LandedOrSplashed` at load time:

- `LandedOrSplashed == true` → `updateMode = IDLE` → transform sticks.
- Otherwise → `updateMode = UPDATE` → on the first physics tick, `updateFromParameters` runs:
  `vessel.SetPosition(referenceBody.position + orbit.pos - localCoM)`
  reading `orbit.pos` from the snapshot's **stale** ORBIT Keplerian elements, captured when the kerbal was still on the parent vessel's ladder, and overwriting the corrected transform with the parent position.

An EVA kerbal committed with any residual `srfSpeed` (walking, jetpack drift, hopping) classifies as FLYING / SUB_ORBITAL / ORBITING at load time and hits the `UPDATE` branch. Only perfectly still kerbals escape the bug.

`KerbalEVA.autoGrabLadderOnStart` (the original suspicion) is trigger-collider driven and cannot fire at 170 m — ruled out.

## Fix

1. Route EVA and breakup-continuous spawns through `VesselSpawner.SpawnAtPosition` (added in #171 for orbital spawns) which rebuilds the `ORBIT` subnode from the endpoint's `lat/lon/alt + velocity` via `orbit.UpdateFromStateVectors` before `ProtoVessel.Load` runs.
2. Extract `OverrideSituationFromTerminalState` helper that generalizes the existing `FLYING to ORBITING` override (#176) to also cover `FLYING to LANDED/SPLASHED` based on `rec.TerminalStateValue`. A walking kerbal at `alt > 0` with low speed classifies as `FLYING`, gets overridden to `LANDED`, `ApplySituationToNode` writes `sit=LANDED`, and the loaded vessel reads `LandedOrSplashed == true` → `updateMode = IDLE`.
3. Add optional `Quaternion? rotation = null` parameter to `SpawnAtPosition` so breakup-continuous spawns preserve `lastPt.rotation` (previously handled via `OverrideSnapshotPosition`).
4. Remove the `!isEva` guard on the bounding-box overlap check in `CheckSpawnCollisions`. Kerbal `ComputeVesselBounds` returns a ~2.5 m cube via `DefaultPartHalfExtent = 1.25f`, adequate clearance with the existing 5 m padding.
5. Add `SpawnCollisionDetector.WalkbackAlongTrajectorySubdivided` — walks backward from the last trajectory point with **1.5 m linear lat/lon/alt sub-steps** until it finds a collision-free candidate. Uses the existing `SurfaceDistance` helper for segment length (flat-Earth approximation, accurate well under 1% at EVA scales). Does **not** modify the existing point-granularity `WalkbackAlongTrajectory` used by `VesselGhoster` for chain-tip spawns — flagged follow-up.
6. Add `VesselSpawner.TryWalkbackForEndOfRecordingSpawn` wrapper calling the subdivided helper with runtime closures, clamping altitude via `ClampAltitudeForLanded` for surface terminals, and on exhaustion setting `SpawnAbandoned + VesselSpawned + Recording.WalkbackExhausted` (new transient field, distinct from KSC-exclusion abandon for diagnostics).
7. Rewrite `CheckSpawnCollisions` to return `(blocked, lat, lon, alt, pos)` so walkback can rewrite the spawn coordinates. Precedence: duplicate-blocker-recovery (#112) runs **first** (a quicksave-loaded duplicate clears with one `RecoverVesselFromFlight`, no walkback needed); walkback runs **second** only if recovery fails or doesn't apply.
8. Walkback triggers **immediately** on first collision — no 5 s timeout like the chain-tip path. End-of-recording spawns have no running ghost to extend during a wait.
9. **Keep** the existing `OverrideSnapshotPosition` calls on the EVA and breakup-continuous paths as defense-in-depth for the `RespawnVessel` fallback: if `SpawnAtPosition` returns 0 the fallback still places the kerbal at the recorded lat/lon/alt on frame 0.

## Scope (partial fix)

This PR closes #264 for the `VesselSpawner.SpawnOrRecoverIfTooClose` **in-flight** spawn path. The following paths still use the older lat/lon/alt override pattern and may partially reproduce the bug — tracked as follow-ups in `docs/dev/todo-and-known-bugs.md`:

- `ParsekFlight.SpawnTreeLeaves` at `ParsekFlight.cs:5926-5991` — scene-load tree-leaf spawn when resuming a saved game with a pending EVA leaf.
- `ParsekKSC.cs:780-847` — KSC-scene spawn path when returning to space center with an EVA recording still pending.
- `VesselGhoster.TryWalkbackSpawn` at `VesselGhoster.cs:421-432` — chain-tip spawn; safe for LANDED tips (`updateMode = IDLE`) but potentially broken for FLYING tips.
- `VesselSpawner.StripEvaLadderState` writes the literal `"idle"` FSM state which is not a valid `KerbalEVA` state name — cosmetic only (the catch block falls through to a real state).
- Triple-correction-layer cleanup: `CorrectUnsafeSnapshotSituation` to `DetermineSituation` to `OverrideSituationFromTerminalState` could collapse into "read corrected `snapshot.sit` first, fall through to classifier only if still FLYING".

## Test plan

- [x] `dotnet test` — **5298 passing**, 0 failing.
- [x] 10 new xUnit tests for `OverrideSituationFromTerminalState` including the Butterfly Rover repro case (`DetermineSituation(alt=5, speed=2, orbitalSpeed=2296) = FLYING` → `OverrideSituationFromTerminalState("FLYING", Landed) = LANDED`).
- [x] 16 new xUnit tests in `EndOfRecordingWalkbackTests.cs` covering the walkback algorithm (null/empty/single-point, off-by-one segment lengths near `stepMeters`, zero-length segments, 100 m segment uses 67 sub-steps, exhaustion + success log assertions, linear interpolation accuracy for lat/lon and altitude, degenerate `stepMeters=0`). All use injected closures, no KSP runtime dependency.
- [x] `dotnet build` — clean, no new warnings.
- [ ] In-game test run (Ctrl+Shift+T): 2 new tests under `EvaSpawnPosition` category. `EvaSpawnAtRecordedEndpoint_NotOnParent` injects a synthetic 20-point trajectory at 5 m spacing walking 100 m due north from the active vessel and asserts the spawned kerbal is within 10 m of the endpoint and >= 50 m from the parent. `EvaSpawnWalkbackOnOverlap` injects a trajectory converging on the parent vessel and asserts walkback moves the spawn off the parent and logs `WalkbackSubdivided: cleared`. Both use a new `InGameKerbalEvaSnapshot` helper (parallel to the test-assembly `VesselSnapshotBuilder`) that populates the full ~23-field `kerbalEVA` PART node so `ProtoVessel.Load` doesn't complain about missing fields. The 5 m point spacing is deliberate — with `stepMeters=1.5` that gives 4 sub-steps per segment, actually exercising subdivision instead of degenerating to point granularity.
- [ ] Manual playtest: EVA from a rover/capsule, walk 100+ m, commit recording, verify spawned kerbal lands at the walked position not on top of the parent.
- [ ] Manual playtest: EVA into orbit, float, commit, verify orbital spawn still works (regression guard for #176).
- [ ] Manual playtest: Breakup-continuous splashdown, verify rotation preserved via the new `SpawnAtPosition` rotation parameter.
- [ ] Manual playtest: Two simultaneous EVAs from the same ship, verify each spawns at its own endpoint.

## Related

- Plan doc: `docs/dev/plans/eva-spawn-position-fix.md` (full design + reviewer pass + reviewer-driven corrections).
- Supersedes the original todo #264 "auto-snap" hypothesis, which decompile proved false.
- Builds on #171 (orbital SpawnAtPosition), #176 (FLYING to ORBITING override), #112 (duplicate-blocker recovery), #231 (LANDED altitude clamp), #224 (breakup-continuous position override).


vl3c:
Review pass addressed in ae49dd8 + merge commit 38889e2 (origin/main merged in):

1. **Walkback scope creep** — CHANGELOG line updated to note the walkback applies to orbital / breakup / EVA / landed spawns, not just EVA. (The merge with main condensed the entry to a one-liner per project convention; the scope note is included there.)
2. **Unbounded n per segment** — added `SpawnCollisionDetector.MaxSubStepsPerSegment = 500` constant and clamped `n` via `Math.Min`. Logs a warn when clamping applies, reporting the effective step size. Normal EVA-scale segments (~67 sub-steps for 100 m / 1.5 m) are well below the cap.
3. **In-game test FixedUpdate margin** — extended the post-spawn wait from 2 `WaitForFixedUpdate` to 3 + `WaitForSeconds(0.1f)`. Defends against a KSP scheduler that defers the first `OrbitDriver` update.
4. **Missing last-point-clear log test** — added `WalkbackSubdivided_LastPointClear_LogsBaseCaseLine` covering the safe-base-case log line.
5. **Tightened bounds sanity assertion** — `EvaSpawnAtRecordedEndpoint_NotOnParent` now asserts `kerbalBounds.size.magnitude < 6.0` (was `< 10.0`). Expected ~4.33 for a 2.5 m cube.
6. **Triple-correction layer inline comment** — `DetermineSituation` now has a `NOTE` block explaining why `snapshot.sit` is ignored and recomputed, with a link to the #264 follow-up entry.

Also merged `origin/main` in — all 5318 tests pass. The original 5298 + 20 new from the main-side commits #272–#275.

PR #200, opened 2026-04-10
Phase 11: Resource Snapshots — detailed plan

vl3c:
## Summary
- Detailed implementation plan for Phase 11 (Resource Snapshots) — prerequisite for Phase 12 (Looped Transport Logistics)
- Five tasks: ExtractResourceManifest, Recording fields + serialization, capture at 14 boundary sites, UI tooltip, ComputeResourceDelta
- Site-by-site analysis of every snapshot capture point with exact placement decisions
- EC excluded at extraction time (fluctuates frame-to-frame, meaningless for logistics)

## Test plan
- [x] Review plan for completeness and correctness before implementation begins

### DIFF
--- a/docs/dev/plans/phase-11-resource-snapshots.md
+++ b/docs/dev/plans/phase-11-resource-snapshots.md
@@ -1,0 +1,515 @@
+# Phase 11: Resource Snapshots — Detailed Plan
+
+## Goal
+
+Make recordings resource-aware so they know what physical resources (LF, Ox, Ore, MonoProp, EC, etc.) a vessel carried at recording start and end. This is the prerequisite for Phase 12 (Looped Transport Logistics) where looped recordings become automated supply routes.
+
+## What we already have
+
+**Vessel snapshots contain full resource state.** Every boundary already calls `VesselSpawner.TryBackupSnapshot()`, which serializes a full ProtoVessel to ConfigNode. Each PART node contains RESOURCE nodes with `amount` and `maxAmount`. The data is already captured — it's just buried inside ConfigNode trees.
+
+**Snapshot capture points already exist** at recording start, stop, chain boundaries (dock/undock/decouple/split/EVA), background recorder finalization, breakup child creation, and optimizer splits.
+
+**Serialization patterns are established.** The `CrewEndStates` Dictionary serialization in `RecordingStore.cs:3037-3099` is the exact pattern to follow: wrapper `AddNode` → child `AddNode("ENTRY")` per key-value pair → batch logging. Both standalone (`ParsekScenario`) and tree (`RecordingTree`) paths follow identical conventions.
+
+---
+
+## T11.1 — ExtractResourceManifest (pure extraction)
+
+### What
+
+A pure static function that walks PART > RESOURCE nodes in a vessel snapshot ConfigNode and returns a summed-by-resource-name dictionary.
+
+### Signature
+
+```csharp
+// In VesselSpawner.cs (co-located with TryBackupSnapshot)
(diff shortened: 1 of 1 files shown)


vl3c:
## Plan Review (fresh Opus agent)

### Blocking

**B1. UI example shows EC but EC is excluded at extraction time.** T11.4 format example includes `EC: 200 → 180 (-20)`, contradicting the T11.1 EC exclusion. Remove or replace with MonoPropellant.

**B2. `ResourceAmount` is a struct — `manifest[name].amount += amount` modifies a copy, not the dict entry.** Must read-modify-write:
```csharp
var ra = manifest[name];
ra.amount += amount;
ra.maxAmount += maxAmount;
manifest[name] = ra;
```
Call this out explicitly in the plan to prevent a silent-zero-accumulation bug during implementation.

**B3. Continuation StartResources override targets the wrong recording.** Plan says to set `contRec.StartResources` in `CommitSegmentCore` after `ApplyPersistenceArtifactsFrom`, but `CommitSegmentCore` operates on the **outgoing** segment being committed, not the incoming continuation. The override must happen on the **next** recorder's `pendingStartResources`, not on the committed segment. Rethink where the boundary-time resources flow to the continuation.

### Worth Fixing

**W1. `CopyStartLocationFrom` exists as a parallel for start fields.** `ApplyPersistenceArtifactsFrom` intentionally excludes start location fields — they have their own copy method. Plan adds `StartResources` to `ApplyPersistenceArtifactsFrom` but StartResources is conceptually a "start field." Document why it breaks the pattern (chain continuations need boundary-time resources, not parent's start).

**W2. `MergeInto` only handles EndResources.** Explicitly state that target.StartResources is intentionally unchanged (represents the earlier start). Prevents someone adding a symmetric line.

**W3. `SplitAtSection` resource handling needs pseudocode.** The prose description is correct but the 200-line method with 14+ numbered steps needs explicit step placement, not just a prose note.

**W4. Phase 12 delivery manifest needs chain-level correlation.** `ComputeResourceDelta` (T11.5) computes whole-recording start-vs-end. Phase 12 needs per-dock/undock boundary deltas. Confirm the data is sufficient — Phase 12 must walk chain structure to find dock/undock segment boundaries.

**W5. Background finalization EndResources is from split-time, not finalization-time.** The snapshot on background debris is from the initial split. EndResources would equal StartResources. Consider just leaving EndResources null for all background-finalized recordings.

**W6. `pendingBoundaryResources` adds hidden state.** Simpler alternative: continuation code reads `capture.EndResources` directly (same object reference until committed). Avoids a new private field.

**W7. Chain of custody for `pendingStartResources` is fragile.** Verify during implementation that `BuildCaptureRecording` sets `capture.StartResources = pendingStartResources` before the capture reaches `ApplyPersistenceArtifactsFrom`.

### Minor

**N1.** Line number references are within 1-2 lines — fine for a plan.

**N2.** `ResourceAmount` struct location should be decided before implementation. Separate file (like `TrajectoryPoint`) or in `Recording.cs`.

**N3.** Round-trip precision tests use different amounts in T11.1 vs T11.2 — unify or explain.

**N4.** Format example uses abbreviated names (`LF`, `Ox`) but decision says "leave as-is" (full names). Pick one.

**N5.** Missing integration test: extract from a `VesselSnapshotBuilder`-produced snapshot, not just hand-built ConfigNodes.

**N6.** Plan says "ignore `flowState`" but doesn't note it in the algorithm. Add a comment to prevent filtering by it.

### EC Exclusion — Correct

Excluding EC is the right call. Also consider excluding **IntakeAir** — it has dynamic maxAmount based on intake area/speed, amounts are meaningless noise, no one transports it. **Ablator** should stay — it's a meaningful consumable. **SolidFuel** stays — it's a real resource on boosters.

PR #223, opened 2026-04-11
Fix ghost thermal visuals: clear emissive, decouple from throttle, merge variant fix (#306, #242c)

vl3c:
## Summary

- **#306 emissive fix**: Ghost engine nozzles permanently glowed red because cloned materials inherited non-zero `_EmissiveColor` from prefabs. Force `coldEmission = Color.black` and clear emissive on clone in both `BuildHeatMaterialStates` and `CollectReentryGlowMaterials`.
- **#306 thermal decoupling**: Removed `ApplyHeatState` calls from `EngineIgnited/Shutdown/Throttle` and `RCSActivated/Stopped` handlers. Ghost thermal glow is now driven purely by `ModuleAnimateHeat` temperature threshold events.
- **#306 threshold adjustment**: Changed heat thresholds to cool <40%, warm 40-80%, hot >80% (was <10%, 33%+, 66%+). Hysteresis gaps at [0.35, 0.40) and [0.75, 0.80).
- **#242c merge**: Variant geometry filtering for engine/RCS FX (fixes Poodle 3-flame regression).

## Test plan

- [x] 5753 unit tests pass
- [x] No log spam from new emissive clear logging
- [x] Ghost engine nozzles do not glow red at rest
- [x] Reentry glow still activates during high-speed atmospheric flight
- [x] Poodle ghost shows 2 flames, not 3
- [ ] RAPIER mode-switch FX still correct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to Parsek are documented here.
 - `#297` Vessel destruction during tree recording no longer orphans continuation data as a standalone recording.
 - `#304` Stock vessel recordings now show resolved names instead of raw `#autoLOC` keys in the UI, timeline, and logs.
 - `#305` Standalone recordings (rovers, planes) now survive revert-to-launch and show the merge dialog instead of being silently discarded.
-- `#306` Ghost engine nozzles no longer glow red permanently -- inherited prefab emissive values are cleared to black at build time.
+- `#306` Ghost engine nozzles no longer glow red permanently -- inherited prefab emissive values are cleared to black at build time. Thermal glow now driven purely by temperature thresholds (cool <40%, warm 40-80%, hot >80%), decoupled from engine/RCS throttle.
 - Landed ghost terrain correction now clamps downward when terrain height decreased since recording (was only clamping upward, leaving ghosts floating above ground).
 - Test runner window no longer spams IMGUI layout exceptions when opened during flight.
 - `T55` FlagEvents and SegmentEvents are now preserved across tree recording splits and flushes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@ All notable changes to Parsek are documented here.
 - `#241` Ghost parts with the base/default color variant (e.g., BlackAndWhite fuel tanks) no longer show the wrong variant texture (was Orange).
 - `#297` Vessel destruction during tree recording no longer orphans continuation data as a standalone recording.
 - `#304` Stock vessel recordings now show resolved names instead of raw `#autoLOC` keys in the UI, timeline, and logs.
-- `#306` Ghost engine nozzles no longer glow red permanently -- inherited prefab emissive values are cleared to black at build time.
 - `#305` Standalone recordings (rovers, planes) now survive revert-to-launch and show the merge dialog instead of being silently discarded.
+- `#306` Ghost engine nozzles no longer glow red permanently -- inherited prefab emissive values are cleared to black at build time.
 - Landed ghost terrain correction now clamps downward when terrain height decreased since recording (was only clamping upward, leaving ghosts floating above ground).
 - Test runner window no longer spams IMGUI layout exceptions when opened during flight.
 - `T55` FlagEvents and SegmentEvents are now preserved across tree recording splits and flushes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to Parsek are documented here.
 - `#241` Ghost parts with the base/default color variant (e.g., BlackAndWhite fuel tanks) no longer show the wrong variant texture (was Orange).
 - `#297` Vessel destruction during tree recording no longer orphans continuation data as a standalone recording.
 - `#304` Stock vessel recordings now show resolved names instead of raw `#autoLOC` keys in the UI, timeline, and logs.
+- `#306` Ghost engine nozzles no longer glow red permanently -- inherited prefab emissive values are cleared to black at build time.
 - `#305` Standalone recordings (rovers, planes) now survive revert-to-launch and show the merge dialog instead of being silently discarded.
 - Landed ghost terrain correction now clamps downward when terrain height decreased since recording (was only clamping upward, leaving ghosts floating above ground).
 - Test runner window no longer spams IMGUI layout exceptions when opened during flight.
--- a/Source/Parsek.Tests/PartEventTests.cs
+++ b/Source/Parsek.Tests/PartEventTests.cs
@@ -2056,7 +2056,7 @@ namespace Parsek.Tests
 
             var evt = FlightRecorder.CheckAnimateHeatTransition(
                 key, 450, "shockConeIntake",
-                normalizedHeat: 0.70f,
+                normalizedHeat: 0.85f,
                 levelMap, ut: 100.0, moduleIndex: 0);
 
             Assert.True(evt.HasValue);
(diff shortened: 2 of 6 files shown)


vl3c (commented):
## Code Review: Fix ghost engine nozzles permanently glowing red (#306)

### 1. Correctness

The root cause analysis is accurate. KSP engine nozzle prefab materials ship with non-zero `_EmissiveColor` baked in because the stock thermal system (`TemperatureGagueSystem`) drives that property at runtime from `part.temperature`. Ghost parts have no temperature simulation, so the inherited value became a permanent baseline glow.

The fix is correct in both dimensions:
- **Material property:** `SetColor(emissiveProperty, Color.black)` immediately after cloning zeroes out the inherited value on the GPU-side material instance.
- **State struct:** `coldEmission = Color.black` (unconditional) and `hotEmission = HeatEmissionColor` (no longer `coldEmission + HeatEmissionColor`) ensures the heat/reentry animation range is [black, hot] rather than [inherited, inherited+hot]. This is the right change -- the old additive formula would have compounded the prefab emissive on top of the intended hot color, which was also a latent bug for the hot end of the spectrum.

No edge cases missed that I can see. `TryGetHeatEmissiveProperty` already returns null for materials without any of the three emissive property names, and the `if (emissiveProperty != null)` guard correctly skips those.

### 2. Consistency

Both `BuildHeatMaterialStates` and `CollectReentryGlowMaterials` receive identical treatment:
- Same `SetColor(emissiveProperty, Color.black)` clear after clone
- Same `coldEmission = Color.black` unconditional assignment
- Same `hotEmission` simplification (drop the `coldEmission +` additive term)

The only asymmetry is that `BuildHeatMaterialStates` gets the `clearedEmissives` counter and log line while `CollectReentryGlowMaterials` does not. This is fine -- `CollectReentryGlowMaterials` covers the whole ghost (every renderer not already managed by heat), so logging per-material clears there would be noisy. The `BuildHeatMaterialStates` log is scoped to a single part and gated behind `Verbose`, which is appropriate.

### 3. Side effects

No risk to non-engine parts. The clear only fires when `emissiveProperty != null`, which requires the material to have one of `_EmissiveColor`, `_EmissionColor`, or `_Emissive`. The vast majority of KSP structural/fuel/aero parts use `_Color` only -- they will not be affected. For the few non-engine parts that do have an emissive property (e.g., docking port lights, landing lights), the stock KSP behavior is the same: those emissives are driven at runtime by their respective modules. Ghost parts lack those modules, so clearing to black is the correct cold state for those too.

### 4. Logging

The new `ParsekLog.Verbose` line is appropriate:
- Level: `Verbose` -- this is a one-shot build-time diagnostic, not per-frame. Correct level per CLAUDE.md conventions.
- Content: includes part name, PID, and count of cleared emissives -- good for diagnosing which parts had inherited values.
- Spam: gated behind `clearedEmissives > 0` and only fires once per part during ghost construction. No spam risk.
- Placement: inside the `trackedMaterials > 0` block, before the return. Correct -- only logs when the method actually produced results.

### 5. Docs

**CHANGELOG:** Single line, user-facing, under 2 sentences. Follows the project convention. One minor observation: #306 is inserted before #305 in the list, breaking numerical order. The existing list is not strictly ordered (e.g., the unnumbered entries are interspersed), so this is cosmetic, but worth noting.

**todo-and-known-bugs.md:** Entry follows the standard template (Observed in / Root cause / Fix / Status). Root cause explanation is precise and matches the code change. Correctly marked as ~~strikethrough~~ for fixed status.

### Verdict

Clean, well-scoped fix. The root cause analysis, code change, and documentation are all aligned. No issues found.

vl3c (commented):
## Code Review: PR #223 -- Fix ghost thermal visuals

### 1. Bug: Remaining throttle-to-heat coupling on RCSThrottle

**GhostPlaybackLogic.cs line 1041** still has `ApplyHeatState(state, evt, HeatLevel.Hot)` inside the `RCSThrottle` case:

```csharp
case PartEventType.RCSThrottle:
    SetRcsEmission(state, evt, evt.value);
    ApplyHeatState(state, evt, HeatLevel.Hot);  // <-- not removed
    break;
```

The PR removed `ApplyHeatState` from `EngineIgnited`, `EngineShutdown`, `EngineThrottle`, `RCSActivated`, and `RCSStopped` -- but missed `RCSThrottle`. This contradicts the stated goal of decoupling thermal animation from engine/RCS throttle. Every RCS throttle change event will force the part to Hot, overriding the temperature-threshold-driven state.

**Action required:** Remove the `ApplyHeatState` call from the `RCSThrottle` case to match the other removals.

### 2. Orphan-engine auto-start still forces Hot (line 785)

The orphan-engine FX auto-start path at line 785 still calls `ApplyHeatState(state, syntheticEvt, HeatLevel.Hot)`. This is for debris boosters that have no engine events in their recording, so they need synthetic ignition. Since these parts have no `ModuleAnimateHeat` events either, forcing Hot here is the only way they get any heat visual at all -- so this one seems intentional and correct. Just flagging for awareness since it is technically a throttle-to-heat coupling that survives this PR.

### 3. Hysteresis logic: correct but has a notable edge case

The new `CheckAnimateHeatTransition` logic is structurally sound. Walking through the zones:

- **>= 0.80**: always Hot (correct)
- **[0.40, 0.80)**: Medium, except Hot stays Hot in [0.75, 0.80) (correct hysteresis)
- **[0.35, 0.40)**: preserves current state (correct hysteresis gap)
- **< 0.35**: always Cold (correct)

One edge case worth noting: if a part is currently Hot and heat drops directly into [0.35, 0.40), `target = current` = Hot, so it stays Hot. This means a part can be "Hot" at 0.36 normalized heat, which is visually incongruent. This only happens if heat drops from >= 0.75 to the [0.35, 0.40) range in a single polling frame (a ~0.40 drop), which is unlikely in practice. Not a bug, just a consequence of the hysteresis design -- acceptable.

### 4. Emissive clearing: thorough and correct

Both `BuildHeatMaterialStates` and `CollectReentryGlowMaterials` now:
- Clear the emissive property to `Color.black` immediately after cloning
- Set `coldEmission = Color.black` unconditionally (no longer reading from the clone)
- Compute `hotEmission = HeatEmissionColor` directly (no longer `coldEmission + HeatEmissionColor`)

This correctly addresses the root cause: prefab materials with baked non-zero `_EmissiveColor` were being inherited as the baseline. The init-to-cold loop at line 664-668 provides belt-and-suspenders safety at spawn time. Good.

### 5. Test coverage

The updated tests properly exercise:
- Cold-to-Medium at 0.45 (above 0.40 threshold)
- Medium-to-Hot at 0.85 (above 0.80 threshold)
- Cold-Medium hysteresis gap at 0.37 (Cold stays Cold)
- Medium in Cold-Medium gap at 0.37 (Medium stays Medium)
- Hot in Medium-Hot gap at 0.77 (Hot stays Hot)
- Hot falls below hysteresis at 0.55 (becomes Medium)

**Missing test:** Hot falling into the Cold-Medium hysteresis gap [0.35, 0.40). Per the logic, Hot at 0.37 would stay Hot (target = current). This is the edge case from point 3 above. A test documenting this behavior would be valuable, even if just to assert the design intent.

**Missing test:** Medium falling below 0.35 to Cold. The existing `MediumToCold` test uses 0.05, which is fine, but there is no test at the exact boundary (0.35 should trigger Cold, 0.349... should trigger Cold).

### 6. Documentation

CHANGELOG entry is accurate and within the 2-sentence guideline. The todo-and-known-bugs entry (#306) is thorough with correct root cause, fix description, and status.

Minor: the CHANGELOG mentions "cool <40%, warm 40-80%, hot >80%" but the hysteresis means the effective boundaries are fuzzier than these clean ranges suggest. The todo entry correctly describes the hysteresis gaps, so this is fine for user-facing docs.

### Summary

One actionable bug (RCSThrottle `ApplyHeatState` not removed), one missing test (Hot in Cold-Medium gap). The emissive fix and the hysteresis logic restructuring are both correct.

PR #241, opened 2026-04-13
fix: trigger debris explosion FX at first destroy event

vl3c:
## Summary
- trigger destroyed debris explosion FX from the first recorded destroy event instead of waiting for recording end
- apply the same earlier trigger in flight and KSC playback paths
- add unit coverage for early debris explosion timing

## Testing
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj


vl3c:
@copilot review

PR #257, opened 2026-04-13
Restore in-flight crash merge dialog flow

vl3c:
Summary: restore the in-flight crash merge dialog flow, keep debris-only survivor settling in flight, and prevent stock FlightResults replay outside FLIGHT. Testing: targeted TreeDestruction, RuntimePolicy, FlightResultsPatch, and QuickloadDiscard test coverage passed.


vl3c:
Checked the successful post-fix run and the current branch code. No code changes were needed, so there is no new commit.

Why the merge-dialog message text changed:
- The dialog body comes from `MergeDialog.ShowTreeDialog()`, which builds `message = $"{tree.TreeName} - {FormatDuration(duration)}"` and then appends the explanatory paragraph only when `spawnCount == 0`.
- Before this follow-up, debris and `SubOrbital` leaves were still counted as `canPersist=True`, so some crash trees reported nonzero `spawnable` counts even though those leaves were not truly spawnable at playback time.
- In the follow-up fix, `MergeDialog.CanPersistVessel()` was changed to reuse `GhostPlaybackLogic.ShouldSpawnAtRecordingEnd(...)`, which makes debris and `SubOrbital` leaves ghost-only by default.
- That means the dialog text now legitimately differs for crash trees: if the old dialog thought some debris leaves were spawnable, it would omit the "No flight branches produced a vessel that can continue flying" note; now it can show that note because `spawnCount` is computed from the real spawn policy.
- The successful fix did not change the dialog title format itself; it changed the spawnability classification feeding the optional explanatory paragraph and the logged `spawnable=` count.

Relevant history:
- `5ea03a3 #286 Show explanatory message in merge dialog when all branches destroyed`
- `8ce87d7 Fix debris-only crash merge ordering regression`

Relevant code/log references on the branch:
- `Source/Parsek/MergeDialog.cs` builds the message from tree name + duration and appends the extra paragraph only for `spawnCount == 0`.
- `Source/Parsek/MergeDialog.cs` now derives `CanPersistVessel()` from `ShouldSpawnAtRecordingEnd(...)`.
- Older logs from 2026-04-10 / 2026-04-12 show debris `terminal=SubOrbital ... canPersist=True`, which inflated `spawnable` and changed the dialog copy.

If needed, I can do a follow-up pass that logs the final dialog body string explicitly at show time so future UI-text differences are visible directly in `KSP.log`.

PR #258, opened 2026-04-13
Fix debris ghost initial replay frame

vl3c:
## Summary
- fix debris ghosts appearing offset on their first visible replay frame by separating the visuals container from the playback root frame
- add runtime appearance logging that records ghost root, part, renderer-bounds, playback-point, and snapshot-frame positions on first visibility
- add focused regression coverage for debris snapshot CoM and root-part frame parsing

## Testing
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter GhostVisualFrameTests
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter GhostPlaybackEngineTests (existing Unity GameObject SecurityException in SpawnGhost_PrimesFreshGhostToCurrentPlaybackUT harness path)


vl3c:
Fresh runtime evidence from logs/2026-04-13_1959_debris-slide-still-broken-after-pr258 changed the root-cause read on the debris slide bug.

Key finding:
- this was not primarily a CoM/root-frame mismatch
- debris ghosts were becoming visible off the breakup branch metadata before their first playable payload sample existed
- e.g. ghost #1 first appeared at playback UT 19.72 while its first playable frame was 20.26, and ghost #7 first appeared at 27.58 while its first playable frame was 28.12
- the old GhostAppearance lines showed activeFrame=none on first visibility while the root was already snapped to the future first point, which explains the visible ahead-of-path spawn and slide-back

Follow-up fix now pushed:
- commit 1b6cb7c delays ghost activation until the first real playable payload frame/checkpoint instead of using the outer semantic branch start
- loop/debris-sync start now uses the same resolved activation UT
- GhostAppearance now logs activationStart / activationLead to prove whether a ghost became visible before its real payload window
- focused regression coverage was added for ExplicitStartUT earlier than the first playable debris frame

Docs:
- added a separate todo note for this as #336 in docs/dev/todo-and-known-bugs.md so it stays distinct from #331 false-alarm sparse-resume truncation

Current branch head also includes the merge from origin/main (e75b504).

PR #298, opened 2026-04-14
Fix #373: log landed ghost clearance fallback instead of silent 0.5 m regression

vl3c:
## Summary

Bug `#373` part 1: `ResolveImmediateLandedGhostClearanceMeters` used to return a hardcoded `0.5` m whenever the distance-aware path could not run (no `CelestialBody`, no `FlightGlobals.ActiveVessel`, or both). That literal matched the legacy pre-`#262` floor, so scene-transition / cold-start paths silently regressed to the old buried-ghost behavior. Now the fallback emits a rate-limited warning naming the ghost index, vessel, body, and reason so the regression surfaces in `KSP.log`.

- New `internal const double ImmediateLandedGhostClearanceFallbackMeters = 0.5` pins the literal.
- New `internal static string ResolveImmediateLandedGhostClearanceFallbackReason(bool hasBody, bool hasActiveVessel)` pure helper classifies the fallback cause (`no-body`, `no-active-vessel`, `no-body-and-no-active-vessel`, or `null` when live).
- `ParsekLog.WarnRateLimited("TerrainCorrect", "landed-ghost-clearance-fallback-{index}-{reason}", ...)` rate-limits per ghost+reason combo.
- `ResolveImmediateLandedGhostClearanceMeters` now takes `int index, string vesselName` so the warning can name the offender; both call sites (`PositionAtPoint`, `PositionAtSurface`) updated.

Bug `#373` part 2 (`PositionAtSurface` cache asymmetry) was **stale at review time**. The follow-up commit `e54f9a8c` that added the `ShouldApplyImmediateSurfacePositionClearance` guard already placed the `state.lastInterpolatedBodyName` / `state.lastInterpolatedAltitude` writes OUTSIDE the `if` block, so both the clearance and non-clearance branches in current HEAD already update the cache unconditionally. Consumers (`WatchModeController` camera/atmo/reentry-FX, `DriveReentryToZero`, `GhostPlaybackEngine.UpdateReentryFx`) read the cache after every surface position, so the symmetric "both update" behavior is the correct one — no code change. The todo entry is annotated so the review comment does not get re-raised.

## Test plan

- [x] `dotnet build` from `Source/Parsek` — clean (0 warnings, 0 errors).
- [x] `dotnet test` from `Source/Parsek.Tests` — 6200 pass, only the known `SpawnGhost_PrimesFreshGhostToCurrentPlaybackUT` failure (bug `#380`).
- [x] 5 new unit tests in `TerrainCorrectorTests` cover all four `(hasBody, hasActiveVessel)` combinations of the extracted reason helper plus the pinned fallback constant.
- [ ] In-game smoke: watch a landed ghost across a scene transition and grep `KSP.log` for `landed-ghost-clearance-fallback` to confirm the fallback path is either never hit (normal) or visible when it is (no longer silent).

Touches bug `#373` in `docs/dev/todo-and-known-bugs.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to Parsek are documented here.
 
 _Unreleased._
 
+### Bug Fixes
+
+- `#373` Landed ghost clearance no longer silently regresses to the legacy 0.5 m floor when the distance-aware resolver cannot run — the fallback now emits a rate-limited warning naming the ghost, recording, and reason so the cold-start / scene-transition path is visible in the log.
+
 ---
 
 ## 0.8.1
--- a/Source/Parsek.Tests/TerrainCorrectorTests.cs
+++ b/Source/Parsek.Tests/TerrainCorrectorTests.cs
@@ -178,6 +178,46 @@ namespace Parsek.Tests
         }
 
         [Fact]
+        public void ResolveImmediateLandedGhostClearanceFallbackReason_AllLive_ReturnsNull()
+        {
+            string reason = ParsekFlight.ResolveImmediateLandedGhostClearanceFallbackReason(
+                hasBody: true, hasActiveVessel: true);
+            Assert.Null(reason);
+        }
+
+        [Fact]
+        public void ResolveImmediateLandedGhostClearanceFallbackReason_NoBody_ReturnsNoBody()
+        {
+            string reason = ParsekFlight.ResolveImmediateLandedGhostClearanceFallbackReason(
+                hasBody: false, hasActiveVessel: true);
+            Assert.Equal("no-body", reason);
+        }
+
+        [Fact]
+        public void ResolveImmediateLandedGhostClearanceFallbackReason_NoActiveVessel_ReturnsNoActiveVessel()
+        {
+            string reason = ParsekFlight.ResolveImmediateLandedGhostClearanceFallbackReason(
+                hasBody: true, hasActiveVessel: false);
+            Assert.Equal("no-active-vessel", reason);
+        }
+
+        [Fact]
+        public void ResolveImmediateLandedGhostClearanceFallbackReason_NeitherLive_ReturnsCombined()
+        {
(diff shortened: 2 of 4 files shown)


vl3c:
**Clean-eyes review — approved as-is** (cannot use `gh pr review --approve` on own-account PR).

**Verified:**
- `ResolveImmediateLandedGhostClearanceFallbackReason(bool, bool)` is truly pure: no Unity deps, no static state, no side effects. All 4 input combinations pinned by tests, plus the `0.5` constant pinned.
- `WarnRateLimited` in `ParsekLog.cs` is genuine time-window rate-limiting (default interval, keyed suppression counter), not call-count — load-bearing for the per-physics-frame failure mode.
- Rate-limit key `landed-ghost-clearance-fallback-{index}-{reason}` matches the established `landed-ghost-{index}` convention at `ParsekFlight.cs:9560/9577` in the same file. The 4 reason suffixes each get their own throttle slot, so no silent shadowing between reasons.
- Null-safe sentinels (`"?"`) used when vesselName or bodyName is empty — warning always carries enough context for the log reader.
- **Second sub-claim in #373 confirmed stale**: read `PositionAtSurface` at `ParsekFlight.cs:9611-9613` — both clearance and non-clearance branches update `state.lastInterpolatedBodyName` / `state.lastInterpolatedAltitude` unconditionally outside the `if` block. Commit `e54f9a8c` moved them out. `todo-and-known-bugs.md` #373 documents the stale premise explicitly so future reviewers don't reopen.
- Both call sites (`PositionAtPoint` line 9418, `PositionAtSurface` line 9603) threaded through with `index` + `traj.VesselName`.
- 5 new tests in `TerrainCorrectorTests.cs` cover all 4 `(hasBody, hasActiveVessel)` permutations plus the constant pin.
- CHANGELOG entry is one line, user-facing — respects the hard rule.
- **Build:** 0 warnings, 0 errors. **Tests:** Failed 1, Passed 6200. The single failure is `SpawnGhost_PrimesFreshGhostToCurrentPlaybackUT` (known #380, ECall security exception, unrelated to this PR).

No follow-ups. Ready to merge.

PR #299, opened 2026-04-14
Add EVA boundary merge round-trip test for bug #371

vl3c:
## Summary

- Bug #371 called for a round-trip test proving PR #248's `NormalizeContinuousEvaBoundaryMerge` is stable across a v3 binary sidecar save/load/optimize/resave/reload cycle now that PR #230 made `TrackSections` the on-disk truth for `RecordingFormatVersion >= 1`.
- New `EvaBoundaryMergeRoundTripTests` file with two xUnit tests:
  - `MergeInto_ContinuousEvaBoundary_V3Roundtrip_PointsAndOrbitSegmentsStable` — builds two EVA-tagged atmo/surface recordings sharing a duplicate boundary frame, saves as `BinaryV3`, loads, calls `RecordingOptimizer.MergeInto` (which triggers `CanMergeContinuousEvaAtmoSurfaceBoundary` → `NormalizeContinuousEvaBoundaryMerge` → `TrimOverlappingSectionFrames` → `TrySyncFlatTrajectoryFromTrackSections`), deep-copies the post-merge `Points` / `OrbitSegments`, re-saves, re-loads, and asserts the second-load trajectory matches the captured snapshot field-by-field. Also pins the duplicate-frame trim (6 total section frames → 5).
  - `CanMergeContinuousEvaAtmoSurfaceBoundary_OrbitalPhasePair_IsNotAllowed` — documents the invariant that the continuous-EVA normalizer rejects orbital phase pairs (`SegmentPhase == "exo"`), so the `Absolute`/`Relative`-only `TrimOverlappingSectionFrames` gate cannot observe an `OrbitalCheckpoint` section via this code path. Answers the todo note about whether EVA + orbital-checkpoint is even possible.
- Test-only PR. No production changes — the optimizer round-trip is already stable.

## Test plan

- [x] `dotnet build` (Parsek) — 0 warnings, 0 errors
- [x] `dotnet build` (Parsek.Tests) — 0 warnings, 0 errors
- [x] `dotnet test --filter "FullyQualifiedName~EvaBoundaryMergeRoundTripTests"` — 2 pass, 0 fail
- [x] `dotnet test` full suite — 6197 pass, 1 known-fail (`SpawnGhost_PrimesFreshGhostToCurrentPlaybackUT`, bug #380, unrelated)

Closes bug #371.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@ All notable changes to Parsek are documented here.
 
 ## 0.8.2
 
-_Unreleased._
+### Tests
+
+- `#371` Added a `MergeInto` continuous-EVA boundary merge round-trip test covering v3 binary sidecar save/load/optimize/resave/reload, plus a companion assertion that the optimizer rejects orbital-phase pairs.
+
 
 ---
 
--- a/Source/Parsek.Tests/EvaBoundaryMergeRoundTripTests.cs
+++ b/Source/Parsek.Tests/EvaBoundaryMergeRoundTripTests.cs
@@ -1,0 +1,493 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using UnityEngine;
+using Xunit;
+
+namespace Parsek.Tests
+{
+    [Collection("Sequential")]
+    public class EvaBoundaryMergeRoundTripTests : IDisposable
+    {
+        private readonly string tempDir;
+
+        public EvaBoundaryMergeRoundTripTests()
+        {
+            RecordingStore.SuppressLogging = true;
+            RecordingStore.ResetForTesting();
+            ParsekLog.SuppressLogging = true;
+
+            tempDir = Path.Combine(Path.GetTempPath(),
+                "parsek-eva-boundary-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(tempDir);
+        }
+
+        public void Dispose()
+        {
+            ParsekLog.ResetTestOverrides();
+            ParsekLog.SuppressLogging = true;
+            RecordingStore.SuppressLogging = true;
+            RecordingStore.ResetForTesting();
+
+            if (Directory.Exists(tempDir))
+            {
+                try { Directory.Delete(tempDir, true); }
+                catch { }
+            }
+        }
+
+
+        // Bug #371: regression test for PR #248 NormalizeContinuousEvaBoundaryMerge.
+        //
+        // PR #248 added boundary-frame trimming to MergeInto continuous-EVA path. The merged
+        // TrackSections are the authoritative payload on disk for RecordingFormatVersion >= 1
+        // (PR #230). This test proves that after the optimizer boundary-merge normalizer runs
+        // on loaded recordings, saving and reloading the target yields byte-identical Points
+        // and OrbitSegments -- i.e. the trimmed sections serialize and round-trip cleanly.
+        //
+        // Flow:
+        //   1. Build two EVA-tagged section-authoritative recordings (atmo + surface, Kerbin).
+        //   2. Save both as binary v3 sidecars.
+        //   3. Load both back.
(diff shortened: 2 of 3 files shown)


vl3c:
Independent clean-eyes review — approved as-is, no follow-up commit.

**File hygiene (flagged hotspot)**: `Source/Parsek.Tests/EvaBoundaryMergeRoundTripTests.cs` reads cleanly end-to-end — no bash-echo escape damage, matched braces, no literal `\n`, no doubled quotes, no stray artifacts. Parses as valid C#.

**Merge-path coverage**: The fixture satisfies every `CanMergeContinuousEvaAtmoSurfaceBoundary` gate (`HasSameEvaIdentity`, common body `Kerbin`, `IsAtmoSurfacePhasePair(atmo, surface)`, `HasContinuousBoundaryTiming` with atmo `EndUT == surface StartUT == t0+10`). The test also asserts `CanMergeContinuousEvaAtmoSurfaceBoundary(target, absorbed) == true` as a precondition, so any future fixture drift fails loudly instead of silently bypassing `NormalizeContinuousEvaBoundaryMerge`.

**API signatures**: `RecordingOptimizer.MergeInto(Recording, Recording)` returns `string`, test ignores return (fine). Both `MergeInto` and `CanMergeContinuousEvaAtmoSurfaceBoundary` are `internal static`; `Parsek/Properties/AssemblyInfo.cs` declares `InternalsVisibleTo("Parsek.Tests")`. `WriteTrajectorySidecar` / `TryProbeTrajectorySidecar` / `DeserializeTrajectorySidecar` / `TrajectorySidecarProbe` / `TrajectorySidecarEncoding.BinaryV3` / `CurrentRecordingFormatVersion = 3` all exist and match.

**BinaryV3 path exercised, not flat fallback**: Recordings built with `RecordingFormatVersion = CurrentRecordingFormatVersion (3)`, so `WriteTrajectorySidecar` takes the `>= 2` branch into `TrajectorySidecarBinary.Write`. Test explicitly asserts `probe.Encoding == BinaryV3` for both initial writes and the merged re-save (three guards against flat fallback). On deserialize, the section-authoritative branch rebuilds `Points` from `TrackSections`, so the trimmed 5-frame payload round-trips through real binary serde.

**Trim math (6→5)**: Atmo `[t0, t0+5, t0+10]`, surface `[t0+10, t0+13, t0+20]`. After `MergeInto` concats sections and `TrimOverlappingSectionFrames` walks them sorted, the surface section's first frame (`ut == previousEndUT`) is dropped. Final: 3 + 2 = 5. Both `totalSectionFrames == 5` and `reloadedMerged.Points.Count == 5` are asserted — the latter confirms `TrySyncFlatTrajectoryFromTrackSections` ran and rebuilt Points from the trimmed sections.

**Field-by-field equality**: `AssertPointEqual` / `AssertOrbitSegmentEqual` compare every serialized scalar (ut/lat/lon/alt/body/velocity xyz/rotation xyzw/funds/science/reputation and all 9 orbit fields) — no reference-equality tautology. Snapshot is deep-copied before the second save, so reload cannot alias in-memory merged state.

**Orbital-phase rejection test**: Not a tautology. Fixture satisfies `HasSameEvaIdentity` and `TryGetCommonRecordingBody` but sets `SegmentPhase = "exo"` on both sides, so only `IsAtmoSurfacePhasePair` returns false — pins exactly the invariant #371 asked to document (orbital-checkpoint sections can never reach `TrimOverlappingSectionFrames` via this flow).

**Build + tests**: `dotnet build` clean; filtered run `Passed: 2`; full suite `Failed: 1, Passed: 6197` — single failure is the known `GhostPlaybackEngineTests.SpawnGhost_PrimesFreshGhostToCurrentPlaybackUT` security-exception (bug #380), unrelated.

**Docs**: CHANGELOG entry is a single one-liner under `0.8.2 / Tests`. `docs/dev/todo-and-known-bugs.md` #371 has strikethrough title, `**Fix:**` paragraph naming both new tests, and `**Status:** ~~Fixed~~`. All three matches HEAD.

LGTM — approved as-is.

PR #307, opened 2026-04-15
fix: career-mode earnings pipeline bundle (#394-#405)

vl3c:
## Summary

Closes a cascade of 11 interlocked career-mode ledger bugs that drained funds to zero after commits, lost accepted contracts on scene transition, pinned science at the starting seed, and zeroed out milestone rewards. Also repairs existing bricked c1/sci1 saves on first load, and fixes one latent dedup-key collision surfaced by #405.

Bugs closed: #394, #395, #396, #397, #398, #400, #401, #402, #403, #404, #405.

### Root causes (three independent pipeline breaks that fed each other)

1. **#397** — `RecordingStore` cleared `PendingScienceSubjects` before `LedgerOrchestrator.OnRecordingCommitted` read them, so `ScienceEarning` actions never entered the ledger and `PatchScience` dragged R&D back to seed every commit.
2. **#405 / #404** — Five `GameStateRecorder` handlers (`OnContractAccepted/Completed/Failed/Cancelled/PartPurchased`) wrote to the event store but skipped the real-time `OnKscSpending` ledger write their KSC peers use. `PatchContracts` then unconditionally wiped `ContractsFinished` and the Offered pool every recalc.
3. **#403 / #400** — `GameStateEventConverter` dropped `FundsChanged/ScienceChanged/ReputationChanged` wholesale. `OnProgressComplete` hardcoded milestone rewards to 0.

### Key design decisions

- **Do NOT re-emit `FundsChanged` as `FundsEarning`.** The drop block in `GameStateEventConverter.cs:121-130` stays intact with an expanded regression-guard comment. Re-emission would double-count against the existing recovery/contract/milestone channels. #394 is instead closed by a commit-time reconciliation diagnostic that WARN-logs any mismatch between the dropped events and the emitted earning actions.
- **#400 milestone capture uses a Harmony postfix on `ProgressNode.AwardProgress`.** Decompile confirmed that `GameVariables.GetProgressFunds/Rep/Science` does NOT exist in stock KSP despite the original todo suggesting it.
- **#397 clear must fire in `NotifyLedgerTreeCommitted` AFTER the foreach (not in `OnRecordingCommitted` tail).** Multi-recording trees must all see the same `PendingScienceSubjects` before the clear. Three upstream call sites (`NotifyLedgerTreeCommitted`, `ChainSegmentManager.CommitSegmentCore:527`, `ParsekFlight.FallbackCommitSplitRecorder`) all wrap the call in `try/finally` to guarantee the clear even on throw. A fault-injection hook (`OnRecordingCommittedFaultInjector`) makes the throw-path invariant directly testable.
- **Latent dedup collision fixed in the same PR.** `GetActionKey` for `FundsSpending` returned `RecordingId ?? ""`; KSC `PartPurchased` actions have `recordingId=null`, so every KSC part purchase collided with every other. Added a `DedupKey` field populated from the event key (part name) and included in the composite key.
- **Save-recovery migrations (#401/#396) ship in this PR, not a follow-up.** c1 and sci1 are bricked right now. `TryRecoverBrokenLedgerOnLoad` runs once on load, epoch-isolated, idempotent via `LedgerHasMatchingAction` guard, covers contract/part events + committed science subjects. Scope deliberately limited to the event types #405 stripped; see the doc comment on `IsRecoverableEventType` for the justification.

### Tests

70 new tests across 11 new test files, all in `[Collection("Sequential")]` where they touch shared static state, all using `ParsekLog.TestSinkForTesting` for log assertions:

- `FundsSpendingDedupKeyTests` (6)
- `GameStateRecorderLedgerTests` (7)
- `PatchContractsPreservationTests` (7)
- `OnContractOfferedStoreTests` (3, incl. IL-scan regression guard)
- `PendingScienceSubjectsClearTests` (5, incl. fault-injection throw path)
- `MilestoneRewardCaptureTests` (7)
- `PatchFundsSanityTests` (7)
- `EarningsReconciliationTests` (8)
- `ScienceSubjectPatchHardeningTests` (5)
- `LedgerRecoveryMigrationTests` (12, incl. epoch isolation + idempotency)
- `VesselCostRecoveryRegressionTests` (3)

Final result: **6272 passed, 1 skipped, 0 failed** (1 pre-existing unrelated skip; 70 new tests = 6202 -> 6272).

### Documentation

- `CHANGELOG.md` — single grouped 2-sentence entry covering all 11 bug numbers, per CLAUDE.md HARD RULE.
- `docs/dev/todo-and-known-bugs.md` — 11 bugs struck through with accurate Fix descriptions; #399 kept open with a note that it's now unblocked.
- `docs/dev/plans/career-earnings-bundle.md` — approved plan (synthesized from Plan agent draft + independent review).
- `docs/dev/plans/career-earnings-bundle-review.md` — plan review that informed the corrections.
- `docs/dev/plans/career-earnings-bundle-implementation-review.md` — code review of the implementation.

## Test plan

- [x] Automated: `cd Source/Parsek.Tests && dotnet test` — expect 6272 pass / 1 skip / 0 fail
- [x] Automated: `cd Source/Parsek && dotnet build -c Release` — expect clean build
- [ ] Smoke (manual, requires KSP): load c1 save — verify `TryRecoverBrokenLedgerOnLoad` logs `synthesized N actions`, verify no `PatchFunds` sanity WARN fires, verify `Funding.Instance.Funds` is non-zero in the top bar
- [ ] Smoke (manual, requires KSP): load sci1 save — verify science subjects repaired, R&D pool reflects ~16.44 recovered science, no `PatchScience` sanity WARN
- [ ] Smoke (manual, requires KSP): fresh career mode — achieve `FirstLaunch` milestone at pad, verify `ProgressRewardPatch` postfix fires, verify Actions window shows non-zero funds/rep reward immediately (tests the recalc-timing concern from implementation review)
- [ ] Smoke (manual, requires KSP): fresh career mode — accept a contract from Mission Control, transition to VAB and back, verify the contract is still accepted (#405+#404 regression)
- [ ] Smoke (manual, requires KSP): fresh career mode — purchase multiple parts in VAB, verify funds balance updates correctly and no parts are silently dropped (dedup-collision regression)

## Known gaps / follow-ups

- Migration covers only the event types #405 stripped. If a future broken save surfaces with stale `CrewHired`/`TechResearched`/`FacilityUpgraded` actions, `IsRecoverableEventType` must be extended — tracked as a known limitation under #401 in todo-and-known-bugs.md.
- Milestone science reward is captured and logged but has no dedicated `GameAction` field (schema only has `MilestoneFundsAwarded`/`MilestoneRepAwarded`). Reconciliation diagnostic catches mismatches; long-term fix is a schema field.
- #399 (`ScienceModule.ComputeTotalSpendings` same-UT walk) stays open — suspect, unverified, distinct root cause, easier to confirm now that the ledger is accurate.


vl3c:
## Follow-up: four regressions caught by Codex review

An independent Codex review of this PR surfaced four real bugs that the original orchestration pipeline (plan → review → implement → review) missed. All four are fixed in the commits pushed above; final `dotnet test` state is **6289 pass / 1 skip / 0 fail** (up from 6272; +17 new regression tests across 5 test files).

### Fix 1 (P1) — Tree commits duplicated pending science per recording

Commit: `d19f3957`

`NotifyLedgerTreeCommitted` re-read `GameStateRecorder.PendingScienceSubjects` on every recording in the tree, so an N-recording tree credited each subject N times. `ScienceModule` summed every copy and over-credited the R&D pool by a factor of N.

The irony: the existing regression test `NotifyLedgerTreeCommitted_MultiRecording_AllRecordingsSeeSubjectsBeforeClear` had actively **locked in the duplication** by asserting each recording produced its own full action set. Rewritten to assert exactly one recording absorbs the batch and the total ledger count equals the seed count.

Fix: `OnRecordingCommitted` gains an `IReadOnlyList<PendingScienceSubject> pendingScienceOverride` parameter. The 3-arg overload (used by single-recording paths in `ChainSegmentManager.CommitSegmentCore` and `ParsekFlight.FallbackCommitSplitRecorder`) still reads from the static list. `NotifyLedgerTreeCommitted` snapshots once, picks an owner recording via new `PickScienceOwnerRecordingId` (highest `EndUT`, ties broken by `ActiveRecordingId`), passes the snapshot to the owner and the empty sentinel `EmptyPendingScience` to siblings, then clears the static list in the same `try/finally`.

New tests: 4 covering owner selection (highest EndUT, tie-break on active, null guard, empty tree) and the rewritten multi-recording exactly-once regression.

### Fix 2 (P1) — `DedupKey` not persisted across save/load

Commit: `8f1edb61`

KSC `PartPurchased` actions dedup on a composite `RecordingId:DedupKey` (added in §F of the original plan to stop collapsing all same-UT purchases). `SerializeFundsSpending`/`DeserializeFundsSpending` only wrote `fundsSpent` + `fundsSpendingSource` — `DedupKey` never hit disk. On reload every KSC purchase collapsed back to `""` and `TryRecoverBrokenLedgerOnLoad` synthesized the debits again on next load.

Fix: `SerializeFundsSpending` writes the field (sparse: only when non-empty, so `VesselBuild` rows stay compact). `DeserializeFundsSpending` reads it back. The class doc comment is updated — previously it literally said "NOT serialized", a comment that was accurate but wrong.

New tests: 2 round-trip tests (non-null key round-trips cleanly; null stays null after the round trip).

### Fix 3 (P1) — Contract accept dropped advance payment

Commit: `2d02bf5e`

`OnContractAccepted` built its detail string from `title + deadline + failFunds + failRep` but never read `contract.FundsAdvance`. `ConvertContractAccepted` had no `funds=` parser either. The downstream schema (`GameAction.AdvanceFunds` field, `FundsModule.cs:340` consumption) was already wired — but the capture side wrote zero, so every KSC contract accept under-credited the ledger by the advance amount. Compounds with #403's funds drain.

Fix: `OnContractAccepted` detail format bumped to v3 with `funds=` field (reads `contract.FundsAdvance`). `ConvertContractAccepted` parses it; v2 strings without the key still parse correctly with advance defaulting to 0.

New tests: 3 converter cases (v3 with advance, v2 backward compat, zero advance edge).

### Fix 4 (P2) — Milestone science reward silently dropped

Commit: `2e8b639a`

The `#400` milestone capture wrote `sci=` into the event detail string, but `GameAction` had no `MilestoneScienceAwarded` field and `ConvertMilestoneAchieved` read only `funds` / `rep`. The `sci=` value was silently dropped at convert time. Milestones with science rewards (`Kerbin/Science`, `Kerbin/Landing`, etc.) produced zero ledger science credit.

Fix (schema + consumption):

- New `GameAction.MilestoneScienceAwarded` field, serialized as `milestoneSciAwarded` in `GAME_ACTION` nodes (backward compat: missing key defaults to 0).
- `ConvertMilestoneAchieved` parses `sci=` from detail.
- `ScienceModule.ProcessMilestoneScienceReward` mirrors the existing `ProcessContractScienceReward` pattern — effective-only, flat reward, not subject-capped. Wired into `ProcessAction`'s switch.
- `ReconcileEarningsWindow` (the #394 cross-check) now counts `MilestoneScienceAwarded` toward `emittedSciDelta` so reconciliation stays correct.
- `GameActionDisplay` shows `+N sci` alongside funds/rep for milestone rows in the Actions window.
- `MilestonesModule` credit log includes `sci` for all milestones going forward.

New tests: 4 converter cases (sci present, all three rewards, partial rewards, backward compat without sci key) + 4 `ScienceModule` cases (effective credits, non-effective skips, zero reward no-op, stacks with existing earnings).

### What my orchestration pipeline missed

For the record so the next iteration gets this right:

- **Fix 1** is the worst. The original plan review explicitly worried about the multi-recording tree case and asked the implementer to assert "both recordings see the list before the clear." That instruction encoded the bug as a regression test, and nobody caught it.
- **Fix 2** is a schema-vs-persistence oversight. I added a field and tested dedup at runtime, but never grepped for the serializer. Codex caught it in one read.
- **Fix 3** is missed scope: the review stopped at "contract accept reaches ledger" and didn't verify payload completeness.
- **Fix 4** was actually flagged by both the implementer (in their open-issues report) and the implementation reviewer — but neither treated it as a blocker. That's a calibration issue with how I scored "merge-blocking" on the first pass.

### Smoke tests still pending

The original manual smoke-test checklist (in the PR body) stands. The follow-ups add no new runtime integrations — milestone science relies on the same `ProgressRewardPatch` Harmony postfix, so the same `FirstLaunch` smoke test exercises both `#400` and Fix 4 simultaneously.

vl3c:
## Fresh Opus review pass → 2 P2 fixes landed in `6acd8a7d`

After the codex-follow-up commits, I asked a fresh Opus agent to re-review the branch cold — specifically looking for the same class of bugs the prior pipeline missed. It found two P2 items, both fixed in `6acd8a7d`:

### P2 #1 — `EnrichPendingMilestoneRewards` dropped `MilestoneScienceAwarded`

`GameStateRecorder.cs:815-831` updated the in-ledger `MilestoneFundsAwarded` and `MilestoneRepAwarded` on the already-written action when the `AwardProgress` postfix fired, but silently skipped the sci field that `2e8b639a` just added. KSC path: `OnProgressComplete → OnKscSpending` writes a zero-reward action, then `AwardProgress → EnrichPendingMilestoneRewards` is supposed to patch it with real values. Funds/rep were patched correctly, sci stayed at zero. This exactly undermined Fix #4 for the KSC path — the case Codex's original review was already worried about.

The existing regression test happened to pass `sci=0`, so the gap was invisible.

Fix: one-line addition at the ledger-update site. Two new tests cover the science path and the all-three-rewards case; the existing test now also asserts `sci=0`.

### P2 #2 — Locale-unsafe float serialization in contract detail strings

`OnContractCompleted`, `OnContractFailed`, `OnContractCancelled` (`GameStateRecorder.cs:230/251/271`) wrote floats via plain string interpolation:
```csharp
var detail = $"title={title};fundsReward={contract.FundsCompletion};...";
```

Plain interpolation uses the system locale. On comma-locale systems (de-DE, fr-FR, ro-RO), fractional values serialize as `"1,5"`, and the IC-based converter parses `0`. This was latent in the pre-PR code but newly load-bearing after `#405` started routing these events through `OnKscSpending` in real time — the project's explicit rule ("ALL float/double serialization must use `ToString('R', InvariantCulture)`") was being violated.

Fix: explicit `InvariantCulture.ToString("R")` for every float field in the three contract handlers. Same pattern applied to `OnPartPurchased` (`cost`) and `OnKerbalAdded` (`hireCost`) — both had the same smell; both are lower-frequency but the fix is free.

### P3 — test tightening

`NotifyLedgerTreeCommitted_SingleRecording_SubjectsReadThenCleared` used `scienceActions >= 2` instead of `== 2`. The looseness was exactly what let the multi-recording duplication slide through the first review pass. Changed to `Assert.Equal(2, scienceActions)`.

### Final state

- **20 commits** ahead of `origin/main`
- **Tests: 6291 pass / 1 skip / 0 fail** (+2 from the new enrichment sci tests)
- Review doc: `docs/dev/plans/career-earnings-bundle-codex-followup-review.md` (on the branch)

The manual smoke-test checklist from the PR body still stands — none of it has been executed.

PR #309, opened 2026-04-15
Fix #406: skip reentry FX build for non-reentry trajectories

vl3c:
## Summary

- Map-view framerate in `test career` collapsed when ~260 showcase ghosts looped simultaneously. `GhostPlaybackEngine.UpdatePlayback` alone blew the 8 ms playback budget (logs: `Playback frame budget exceeded: 12.3ms (259 ghosts, warp: 1x) | suppressed=106`) *before* map OnGUI ran, and `PauseAllGhostAudio: paused 265 primary` confirmed the active fleet size.
- Root cause: `TryPopulateGhostVisuals` called `GhostVisualBuilder.TryBuildReentryFx` unconditionally on every ghost build, and loop cycle boundaries destroy-and-respawn ghosts (`DestroyGhost(reason: "loop cycle boundary")` → `SpawnGhost`). For stationary part showcases, the mesh-combine + `ParticleSystem` + glow material clones were pure waste — they can never produce reentry visuals because `ComputeReentryIntensity` gates on Mach ≥ 2.5 and density ≥ 0.0015 kg/m³.
- Fix: gate the build behind new `TrajectoryMath.HasReentryPotential(IPlaybackTrajectory)` — returns `true` iff the trajectory has any orbit segments or any point has velocity magnitude ≥ `ReentryPotentialSpeedFloor` (400 m/s, well under Mach 1.5 on any stock body). When skipped, `reentryFxInfo` stays `null`; all downstream paths already null-guard (same pattern `ParsekKSC.cs:519` uses for flag ghosts). Matching gate in the `ParsekFlight` preview path.

## Evidence

Logs `logs/2026-04-15_2034_showcase-loop-perf/`:

```
[LOG 20:30:53.058] [Parsek][WARN][Diagnostics] Playback frame budget exceeded: 12.3ms (259 ghosts, warp: 1x) | suppressed=106
[LOG 20:30:53.431] [Parsek][VERBOSE][ReentryFx] combined 7 meshes into emission shape (4346 verts) for ghost #1 | suppressed=839
[LOG 20:30:53.670] [Parsek][VERBOSE][GhostAudio] PauseAllGhostAudio: paused 265 primary + 0 overlap ghost(s)
```

The `suppressed=106` means 106 prior budget-exceedance warnings were rate-limited in a 30 s window — the budget was blown on most frames, not a one-off spike.

## Implementation notes

- `TrajectoryMath.HasReentryPotential` is pure static, takes `IPlaybackTrajectory`, O(n) in point count, called once per ghost build.
- Speed floor 400 m/s chosen because Mach 2.5 thermal FX requires ~675 m/s even in thin Kerbin air; 400 m/s is well under Mach 1.5 on every stock body, so the gate cannot hide any real reentry heating.
- Orbit segments always pass the gate — de-orbit happens at ~2300 m/s regardless of the point sampling density.
- New session counters `DiagnosticsState.health.reentryFxBuildsThisSession` / `reentryFxSkippedThisSession` make the gate visible in live diagnostics.
- Rate-limited `Verbose` log line on each skip identifies the ghost by index, vessel name, and reason.
- Follow-ups not in this PR: reusing ghost `GameObject` across loop cycles instead of destroy/rebuild (would eliminate the remaining per-cycle material/audio clone cost); gating hidden-tier prewarm by reentry potential too.

## Test plan

- [x] `dotnet test --filter "FullyQualifiedName~ReentryPotentialTests"` — 12 new unit tests pass (null / empty / stationary / EVA / 200 m/s Flea / just-below-floor / at-floor / fast suborbital point / diagonal magnitude / orbit-segment-only / speed floor constant).
- [x] `dotnet test` full suite — 6240 passed, 1 skipped (pre-existing).
- [x] `dotnet build` — deployed DLL verified to contain the `Skipped reentry FX build for ghost` UTF-16 string.
- [ ] Manual playtest: load `test career`, enter flight + map view on a vessel surrounded by the Part Showcase trees, confirm FPS no longer tanks and that `reentryFxSkippedThisSession` counter climbs while `reentryFxBuildsThisSession` stays low.
- [ ] Manual playtest: confirm a genuine suborbital or orbital ghost *still* shows reentry glow/fire during descent (e.g., Suborbital Arc, Orbit-1, Kerbin Ascent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to Parsek are documented here.
 
 ### Bug Fixes
 
+- `#406` Map-view framerate with many looping showcase ghosts no longer collapses — stationary and slow recordings skip the reentry FX build that was being thrown away and rebuilt on every loop-cycle boundary.
 - `#362` Terminal crash-end decouple fragments (parachutes, heat shields, late shrapnel) now become proper debris branches at the very end of a recording instead of being silently dropped when the parent vessel is already in its destruction frame.
 - `#370` Hardened the group Watch button log lines against a latent `IndexOutOfRangeException` if `ResolveEffectiveWatchTargetIndex` ever returns `-1` while the click reaches the handler.
 - `#373` Landed ghost clearance no longer silently regresses to the legacy 0.5 m floor when the distance-aware resolver cannot run — the fallback now emits a rate-limited warning naming the ghost, recording, and reason so the cold-start / scene-transition path is visible in the log.
--- a/Source/Parsek.Tests/ReentryPotentialTests.cs
+++ b/Source/Parsek.Tests/ReentryPotentialTests.cs
@@ -204,5 +204,52 @@ namespace Parsek.Tests
             // guaranteeing no false negatives on reentry heating.
             Assert.Equal(400f, TrajectoryMath.ReentryPotentialSpeedFloor);
         }
+
+        [Fact]
+        public void NanVelocityComponent_DoesNotThrow_ReturnsFalse()
+        {
+            // Malformed point with a NaN component — sqrMagnitude yields NaN and
+            // `NaN >= floorSq` is false, so the gate conservatively returns false
+            // without crashing. Protects against any upstream bug that leaks a
+            // degenerate velocity vector into the trajectory.
+            var traj = new MockTrajectory();
+            traj.Points.Add(new TrajectoryPoint
+            {
+                ut = 0, bodyName = "Kerbin",
+                velocity = new Vector3(float.NaN, 0f, 0f),
+                rotation = Quaternion.identity
+            });
+            bool result = false;
+            var ex = Record.Exception(() => result = TrajectoryMath.HasReentryPotential(traj));
+            Assert.Null(ex);
+            Assert.False(result);
+        }
+
+        [Fact]
+        public void InfinityVelocityComponent_ReturnsTrue()
+        {
+            // Infinity passes the floor comparison (Infinity >= floorSq), so the gate
+            // plays it safe and builds reentry FX. Again, this is not expected in
+            // production — the test just locks in the behavior so a later refactor
+            // that adds input sanitization is a conscious choice, not an accident.
+            var traj = new MockTrajectory();
+            traj.Points.Add(new TrajectoryPoint
+            {
+                ut = 0, bodyName = "Kerbin",
+                velocity = new Vector3(float.PositiveInfinity, 0f, 0f),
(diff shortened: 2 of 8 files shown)


vl3c:
## Opus review pass — verdict: ship

I ran an independent Opus code review against this PR. Verdict: **ship**. The reviewer verified the null-safety invariant holds across all 10+ downstream reentry-FX consumer sites, confirmed the gate only runs at spawn time (not per frame), traced the loop-cycle rebuild path through `DestroyGhost` / `SpawnGhost` / `ClearLoadedVisualReferences` for stale-state carry-over, and cross-checked the 400 m/s speed floor against Kerbin/Eve/Duna/Laythe speed-of-sound values.

### Review findings verified
- Gate math and constant (`TrajectoryMath.cs:966`), `IPlaybackTrajectory.HasOrbitSegments` wiring, velocity-capture frame (`FlightRecorder` Krakensbane-corrected `rb_velocityD + GetFrameVelocity()`).
- All reentry-FX consumers null-guard: `GhostPlaybackEngine.UpdateReentryFx` (:1395), explosion-length fallback (:2758), `GhostPlaybackLogic.ResetReentryFx` (:2553), `GhostVisualBuilder.RebuildReentryMeshes` (:6311), `DestroyReentryFxResources` (:2595). `ParsekKSC.cs:519` already established the `reentryFxInfo = null` pattern for flag ghosts.
- Preview parity between engine and `ParsekFlight` paths.
- CHANGELOG / todo numbering (#406 free) / diagnostics counter registration / post-v0.8.0 risk-area scope all clean.

### Non-blocking follow-ups addressed in 3a9cd9ce
1. `HasReentryPotential` XML doc now records that `TrajectoryPoint.velocity` is Krakensbane-corrected inertial (body-co-rotating frame) and that NaN components fail the `>=` comparison harmlessly — closes the documentation hazard the reviewer flagged on the 400 m/s floor.
2. Three new unit tests (NaN velocity component / +Infinity / `Points == null`) cover the input-sanitization gap flagged as test-coverage concern #9. `ReentryPotentialTests` now has 15 cases; full suite 6243 pass.
3. `DiagnosticsComputation` session health report now surfaces `reentryFx built X skipped Y` alongside the existing `builds / destroys` counters, making the gate visible in the live diagnostics stream without needing verbose per-skip log lines.

Items the reviewer explicitly listed as deferred future work (reuse ghost GameObject across loop cycles, gate hidden-tier prewarm by reentry potential) remain out of scope — those would take this from a bug fix into an architectural change.

PR #336, opened 2026-04-17
#432: Gloops ghost-only recordings no longer leak into the career ledger

vl3c:
## Summary

- Apply-side fix: `CreateKerbalAssignmentActions` and `CreateVesselCostActions` in `LedgerOrchestrator` early-return on `rec.IsGhostOnly`, closing the `MigrateKerbalAssignments` leak that reserved a crewed Gloops recording's kerbals for the loop duration.
- Belt-and-braces `FilterOutGhostOnlyActions` pre-pass in `RecalculateAndPatch` drops any action tagged to a ghost-only recording, with an Info log that fires only when something was actually filtered. Empty-RecordingId seed / KSC-forwarded actions are preserved.
- Self-heal: pre-fix saves with stale `KerbalAssignment` rows on a Gloops recording get rewritten to an empty set on next `OnKspLoad` via the existing `MigrateKerbalAssignments` → `ReplaceActionsForRecording` flow.

## Design decisions

No capture-side guards. Per #431's tagging, events during a Gloops window belong to the parallel normal recording (via `GetActiveRecordingIdForTagging`) or to between-mission career state (empty tag). Gloops never owns events. This is also consistent with the planned extraction of Gloops as a standalone mod (see `docs/dev/gloops-recorder-design.md`) — career-ledger concerns stay on the Parsek side of the future boundary.

Plan doc: `docs/dev/plans/fix-432-gloops-no-events.md`. Went through a clean-Opus plan review; P1 fixes applied before implementation (stale `:9221` citation removed, `ExtractCrewFromRecording` source clarified to `GhostVisualSnapshot ?? VesselSnapshot`, `PopulateUnpopulatedCrewEndStates` intentional-no-guard documented, `PurgeEventsForRecordings` line number corrected, test 5 strengthened, empty-tag preservation audit added).

## Test plan

- [x] `dotnet build` — clean (0 warnings, 0 errors).
- [x] `dotnet test` — full suite: 6767 passed, 0 failed, 1 skipped (pre-existing skip).
- [x] New `Source/Parsek.Tests/GloopsEventSuppressionTests.cs` — 10 tests covering both action-creation guards, filter helper (drops + preserves), resolver invariant (`GetActiveRecordingIdForTagging` never returns a Gloops id), purge mechanism (forced Gloops-tagged event), `RecalculateAndPatch` filter-log signal (fires / doesn't), `OnKspLoad` self-heal for stale rows.
- [ ] Playtest recipe: launch a crewed plane on a career save, Start Gloops, fly ~30s, Stop. Open Career State / Kerbals — the plane's kerbals are not listed as reserved for the Gloops recording's duration. `KSP.log` contains `"CreateKerbalAssignmentActions: recording '<id>' is ghost-only — skipping"`.
- [ ] Self-heal recipe: quickload an older save where a Gloops recording already has stale `KerbalAssignment` rows, confirm `MigrateKerbalAssignments` rewrites them on first load.

## Follow-ups

- #435 (new, aspirational) — multi-recording Gloops trees (main + debris + crew children, all `IsGhostOnly`, no vessel spawn). The per-recording `rec.IsGhostOnly` filter this PR adds handles the tree case automatically when #435 lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ All notable changes to Parsek are documented here.
 - `#434` KSP's stock crash/mission report (with its Revert buttons) now shows first on vessel destruction; Parsek's merge dialog no longer pre-empts it. Revert to Launch and Revert to VAB/SPH soft-clear the pending recording — sidecar files and captured events stay on disk so a flight quicksave can still be F9'd back into, while the bumped milestone epoch keeps reverted events out of the current ledger.
 - `#434` Fixed a NullReferenceException in `RevertDetector.Subscribe` that aborted `ParsekScenario.OnLoad` before the merge-dialog dispatch path could run, so going back to Space Center after a flight silently auto-committed the pending tree instead of showing the merge/discard dialog.
 - `#434` Revert to Launch no longer accidentally deletes the in-flight recording's files, so a flight F5 quicksave can still be F9'd back in after reverting. Stale science captured between the launch quicksave and the revert is also now cleared even when no recording was yet stashed.
+- `#432` Gloops (ghost-only) recordings no longer leak kerbal assignments or vessel costs into the career ledger — they are purely visual.
 - `#431` Discarding a recording now reverses every career effect it captured — contracts, tech, crew changes, milestones, and resource deltas are all purged, including events the flush-on-save path had already bundled into a milestone.
 - `#416` New career no longer starts with zero funds.
 - `#416` Crashed-vessel recordings now keep their R (rewind) button.
--- a/Source/Parsek.Tests/GloopsEventSuppressionTests.cs
+++ b/Source/Parsek.Tests/GloopsEventSuppressionTests.cs
@@ -171,6 +171,34 @@ namespace Parsek.Tests
         }
 
         [Fact]
+        public void FilterOutGhostOnlyActions_PreservesMigratedOldSaveActionsWithEmptyTag()
+        {
+            // MigrateOldSaveEvents (LedgerOrchestrator.cs:803) converts pre-ledger
+            // save events into GameActions with null/empty RecordingId — documented
+            // intentional tradeoff: we can't reliably map them to specific recordings.
+            // The filter must keep these alongside the legitimate seed-action empty tags.
+            var ghostOnly = new Recording { RecordingId = "gloops-old", IsGhostOnly = true };
+            RecordingStore.AddRecordingWithTreeForTesting(ghostOnly);
+
+            var actions = new List<GameAction>
+            {
+                // Simulated MigrateOldSaveEvents output: a FundsSpending with null tag.
+                new GameAction { RecordingId = null, Type = GameActionType.FundsSpending, FundsSpent = 1000f },
+                // Simulated InitialFunds seed.
+                new GameAction { RecordingId = "", Type = GameActionType.FundsInitial },
+                // The actually-ghost-only-tagged row that must be dropped.
+                new GameAction { RecordingId = "gloops-old", Type = GameActionType.KerbalAssignment },
+            };
+
+            var filtered = LedgerOrchestrator.FilterOutGhostOnlyActions(actions);
+
+            Assert.Equal(2, filtered.Count);
+            Assert.Contains(filtered, a => a.Type == GameActionType.FundsSpending && a.RecordingId == null);
+            Assert.Contains(filtered, a => a.Type == GameActionType.FundsInitial);
+            Assert.DoesNotContain(filtered, a => a.RecordingId == "gloops-old");
+        }
+
+        [Fact]
         public void FilterOutGhostOnlyActions_NoGhostOnlyRecordings_ReturnsInputUnchanged()
         {
             var normal = new Recording { RecordingId = "normal-only", IsGhostOnly = false };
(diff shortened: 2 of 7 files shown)


vl3c:
### Review feedback addressed

**P0 — rebase:** rebased onto current `origin/main`, picking up the `#434` review fix-tail from PR #335. The previous branch base predated those 4 commits. Force-pushed.

**P1 fixes:**

- **Test 5 (resolver invariant)** — was tautological (`Instance?.activeTree?.ActiveRecordingId ?? ""` trivially returns `""` in xUnit). Now seeds *both* a committed Gloops recording and a committed normal recording, and asserts across 10 iterations that the resolver returns `""` or the normal recording's id — never the Gloops id. This pins down that the empty return is due to the resolver honoring the active tree, not "nothing was committed at all."

- **Test 6 (purge mechanism)** — now exercises the orphan-contract-snapshot branch too. Adds a `ContractSnapshot` alongside the forced-tag `ContractAccepted`, asserts both removed after purge. (Milestone-events branch is already covered indirectly via `MilestoneStore.PurgeTaggedEvents` in #431's test suite.)

- **Test 7 (filter log idempotency)** — now includes a second `RecalculateAndPatch` call. The walk filter fires again on each recalc because it does not mutate `Ledger.Actions`; the self-heal path (`OnKspLoad` → `MigrateKerbalAssignments` → `ReplaceActionsForRecording`) is what rewrites stale rows. Clarifies the separation for future readers.

- **New test `FilterOutGhostOnlyActions_PreservesMigratedOldSaveActionsWithEmptyTag`** — documents the intentional tradeoff that `MigrateOldSaveEvents` output (null `RecordingId`) survives the filter alongside legitimate seed-action empty tags. Can't tell post-migration what recording a null-tag row originated from, and preserving them is the safer choice.

- **Stale `:9221` citation in `todo-and-known-bugs.md` #435** — corrected from `ParsekFlight.cs:9221` (which is `DeleteGhostOnlyRecording`'s delete-button guard) to `GhostPlaybackLogic.cs:3001` (the actual `ShouldSpawnAtRecordingEnd` `IsGhostOnly` check).

- **`CreateKerbalAssignmentActions` guard** — now notes the interaction with `PopulateUnpopulatedCrewEndStates` (the safety-net pass that still runs for Gloops recordings, for the Kerbals-window per-recording-fates view). End-state population continues; only the ledger-action emission is suppressed.

**P2s declined** — em-dash usage matches the existing repo convention in log strings (`feedback_no_special_chars_md` is scoped to `.md` files per the memory).

`dotnet test` — 6778 passed (added 1 test case: 11 in `GloopsEventSuppressionTests`), 0 failed, 1 pre-existing skip.

vl3c:
### Review round 2 addressed

**Finding 1 (the real one):** the previous `FilterOutGhostOnlyActions` only filtered the walk-time copy passed to `Recalculate`. `Ledger.Actions` itself kept stale ghost-only rows, and raw-ledger consumers — `TimelineBuilder` (`Timeline/TimelineBuilder.cs:21` reads `ledgerActions` directly), career-state views, the ledger file on save — would still show them. The CHANGELOG claim "no longer leak vessel costs into the career ledger" was therefore only half-true.

**Fix:** replaced the filter with `PurgeGhostOnlyActionsFromLedger` that **mutates `Ledger.Actions` in place** via a new `Ledger.RemoveActionsForRecording(string)` helper. Runs at the top of `RecalculateAndPatch`. Cross-type (unlike `ReplaceActionsForRecording` which keys on `(type, id)`) — catches every stale row a pre-fix save could have: `KerbalAssignment`, `FundsSpending`, `FundsEarning`, etc.

Tests rewritten to match:
- `PurgeGhostOnlyActionsFromLedger_DropsGhostOnlyRowsAndMutatesLedger` — asserts `Ledger.Actions` itself is mutated.
- `PurgeGhostOnlyActionsFromLedger_PreservesMigratedOldSaveActionsWithNullTag` — documents intentional preservation of `MigrateOldSaveEvents` output.
- `PurgeGhostOnlyActionsFromLedger_NoGhostOnlyRecordings_Noop` — clean no-op path.
- `PurgeGhostOnlyActionsFromLedger_RemovesAllTypesForGhostOnlyRecording` — pins cross-type behavior vs `ReplaceActionsForRecording`'s per-type semantics.
- `RecalculateAndPatch_PurgesGhostOnlyActions_AndSecondCallIsNoop` — first call removes rows and logs; second call is silent (ledger is already clean). Contrast with the previous walk-filter design where the log would re-fire every recalc.

**Finding 2 (the tautological test):** dropped `GetActiveRecordingIdForTagging_WithGloopsAndNormalCommitted_NeverReturnsGloopsId`. The resolver is `Instance?.activeTree?.ActiveRecordingId ?? ""` and `Instance` is null in xUnit, so the assertion was trivially satisfied. The "Gloops never owns events" invariant is architectural (`CommitGloopsRecording` never touches `activeTree`), documented in the plan, audited by code review — not unit-testable without a full ParsekFlight MonoBehaviour harness.

**Plan doc + `#432` status note** updated to reflect purge-not-filter design. CHANGELOG unchanged (the one-liner is now fully accurate).

**Test results:** 6778 passed, 0 failed, 1 pre-existing skip. 11 tests in `GloopsEventSuppressionTests.cs`.

PR #500, opened 2026-04-23
Harden recovery and unowned contract ledger events

vl3c:
## Summary
- Defer recovery funds callbacks until the matching FundsChanged(VesselRecovery) event arrives, preserving distinct same-name callbacks inside the UT epsilon.
- Forward unowned contract lifecycle events only when Emit left no recording id and no live recorder can still own the event.
- Close #552 and #553 in the 0.8.3 docs.

## Validation
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName~GameStateRecorderLedgerTests|FullyQualifiedName~DiscardFateTests: 47 passed
- xhigh follow-up review: no findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,11 +115,15 @@ All notable changes to Parsek are documented here.
 
 ### Bug Fixes
 
+- `#552` Vessel recovery funds now tolerate stock firing `onVesselRecovered` before the paired `FundsChanged(VesselRecovery)` event. Parsek defers the recovery request and pairs it when the funds event arrives, eliminating false missing-payout warnings without fabricating zero recoveries.
+- `#553` Untagged contract lifecycle events now forward directly to the ledger even in FLIGHT, so launch-site contract accept/complete events that occur before any Parsek recording owner exists do not get stranded only in `GameStateStore`. Tagged FLIGHT teardown events remain protected by the non-empty recording tag gate.
 - `#557` Initial science and reputation seeds now prefer captured game-state baselines (including legitimate zero values) over live KSP singleton balances. A zero seed is now authoritative instead of being upgraded later from future live state, so rewind/cutoff recalculations no longer turn post-launch science or reputation into UT0 budget.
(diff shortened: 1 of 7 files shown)


vl3c:
Review fixes applied:

- **Blocker**: Added `FlushStalePendingRecoveryFunds(reason)` that evicts unclaimed `pendingRecoveryFunds` entries on lifecycle boundaries (save load, rewind end, every scene switch) with WARN listing each evicted entry. Safety-net `PendingRecoveryFundsStaleThreshold=5` WARN trips if the queue grows past 5 without draining.
- **Major (#553 scope)**: `ShouldForwardDirectLedgerEvent(evt.recordingId, HasLiveRecorder())` now gates every direct-ledger handler: `TechResearched`, `PartPurchased`, `CrewHired`, `MilestoneAchieved` (both paths), `StrategyActivated`, `StrategyDeactivated`, and `FacilityUpgraded`. The contract handlers were already on the new predicate; the other seven sites are now aligned.
- **Major (recovery pairing)**: `OnRecoveryFundsEventRecorded` now uses a two-tier policy - prefer pending requests whose `VesselName` matches `evt.detail`, fall back to nearest-UT. Ties after the name-match filter log a WARN listing tied candidates.
- **Minor (coverage)**: Added a regression pinning that handlers pass `evt.recordingId` (stamped at Emit time) to the predicate rather than re-reading `ResolveCurrentRecordingTag()` - verified by constructing the asymmetric case where the two disagree.
- **Minor (Initialize)**: Confirmed `Initialize()` already short-circuits via `if (initialized) return;`; added a one-line comment at the `OnRecoveryFundsEventRecorded` call site. Added a cross-ref comment on the `consumedRecoveryEventKeys` field pointing to its sole writer (`TryAddVesselRecoveryFundsAction`).

Tests: 14 new xUnit cases across `GameStateRecorderLedgerTests` (6 cases) and `DiscardFateTests` (8 cases). 8315/8315 pass (47 -> 61 in the two affected classes).

Rebased onto `origin/main` (CHANGELOG + todo-and-known-bugs conflicts resolved by keeping both the `#552`/`#553` and `#557`/`#558` entries with expanded wording to match the new scope).

Ready for re-review.

PR #502, opened 2026-04-23
Restore tech availability on rewind

vl3c:
## Summary
- Build a target R&D tech set from the selected baseline plus affordable science spend actions at the cutoff.
- Patch KSP proto-tech availability so future tech locks again after rewind and rehydrates bypass-entry-purchase parts when tech becomes available again.
- Close #551 in the 0.8.3 docs.

## Validation
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName~KspStatePatcherTests: 22 passed
- xhigh follow-up review: no findings


vl3c:
Review fixes applied and pushed (rebased on current origin/main, then one follow-up commit 81b231e):

- Bug-number collision with #551 resolved: renumbered to #559 in CHANGELOG.md and docs/dev/todo-and-known-bugs.md (559 was the next free slot after 551/557/558).
- `LedgerOrchestrator.RecalculateAndPatch` now only builds and passes `targetTechIds` when `utCutoff.HasValue` (rewind path). Non-rewind calls pass null so `PatchTechTree` short-circuits via its existing null-target guard - live tech unlocks after the latest baseline are no longer clobbered on every commit / KSC action / load tick.
- `EnsureAvailableProtoTechNode` rehydrates bypass-entry-purchase parts unconditionally when bypass is on, so stale/partial `partsPurchased` lists self-heal. Reference + name dedup in `AddPurchasedPartsForTech` keeps re-running safe.
- New diagnostic logging:
  - `GetProtoTechNodesDictionary` emits a one-shot `ParsekLog.Warn` identifying the failed field when the reflection hit returns null (static bool guard, cleared by `ResetForTesting`).
  - "missing targets" log now also emits a `ParsekLog.Verbose` listing the first ~10 missing tech ids.
  - `PatchTechTree: available=...` Info log now includes `utCutoff` and `baselineUt` context.
- Added `KspStatePatcherTests` log-assertion coverage: null target -> Verbose skip, null R&D singleton -> Verbose skip, one-shot reflection-failure warn (fires once, refires after `ResetForTesting`).
- CHANGELOG #559 bullet trimmed to a single line / single sentence per project style.
- Cutoff inclusivity (step 8): `BuildTargetTechIdsForPatch` uses `UT > utCutoff.Value` as the skip predicate, matching the exclusive-upper skip semantics in `RecalculationEngine.Recalculate`, `LedgerOrchestrator` projection walks, and sibling affordability filters. No change needed.

All 8311 xUnit tests pass. Deployed DLL verified to contain the new UTF-16 log strings. Ready for re-review.

PR #503, opened 2026-04-23
Fix KSC merge source vessel adoption

vl3c:
## Summary
- prevent KSC recording-end materialization from spawning a duplicate when the source vessel still exists in the save
- adopt the source vessel persistent id as the resume target so recording resume can continue on the original craft
- document #550 and add focused regression coverage

## Regression source
KSC spawn-at-end from #99 checked only whether Parsek had already spawned a copy, not whether the original source vessel still existed in the save. That allowed a merge-return flow to spawn a duplicate real vessel on top of the original.

## Validation
- dotnet test Source\Parsek.Tests\Parsek.Tests.csproj --no-restore --filter "FullyQualifiedName~KscSpawnTests|FullyQualifiedName~VesselSwitchTreeTests" --verbosity minimal


vl3c:
Superseded by #505, which includes the original KSC source-vessel adoption fix plus the broader spawn-invariant hardening follow-up.

PR #511, opened 2026-04-23
Rewind-to-Staging button surfaces on non-crashed booster slots

vl3c:
## Summary

- Post-v0.8.3 playtest finding: clicking Rewind on "Kerbal X Probe" (a booster that split at staging and was still in orbit) fell through to the legacy tree-root launch-rewind dialog because `EffectiveState.IsUnfinishedFlight` gated the new Rewind-to-Staging button on `TerminalKind.Crashed` per design §7.31.
- `UI/RecordingsTableUI.ResolveUnfinishedFlightRewindRoute` now consults the broader `EffectiveState.IsRewindableSlot`: any committed-visible recording whose parent BranchPoint carries a `RewindPoint` exposes the `Rewind` button — InFlight / Orbiting / SubOrbital / Landed / Splashed / Destroyed all qualify. Unfinished Flights virtual group membership, hide refusal, and drag-into-user-group rejection remain narrow on `IsUnfinishedFlight`.
- Design doc §7.31 rewritten to reflect the lifted limitation. CHANGELOG + todo doc updated.

## Player-facing effect

Before: the `Rewind` button on a Rewind-Point-backed staging child only showed up when that child **crashed**. A booster that staged cleanly and was still coasting, had reached orbit, went suborbital, or landed safely saw only the legacy `R` button, which rewinds all the way back to the parent tree's launch — not what the player wanted.

After: any booster that split off at a multi-controllable staging / undock / EVA event gets a `Rewind` button on its row regardless of how it ended. Clicking it reloads the staging quicksave directly into FLIGHT on that vessel, so you can re-fly the booster back to the pad without rewinding the rest of your mission.

The "Unfinished Flights" virtual group keeps its original behaviour — it still only surfaces crashed slots, because that's the "I have regret to fix" list.

## Test plan

- [x] `dotnet test` — 8374 / 8374 pass (5 new `IsRewindableSlot_*` cases, 4 new `TryResolveUnfinishedFlightRewindPoint_{Landed,Orbiting,SubOrbital,InFlight}ChildResolvesToStagingSlot` cases plus `NoParentBranchPointDoesNotResolve`, `ParentBpWithoutRpDoesNotResolve`, `NotCommittedMergeStateDoesNotResolve` negatives; the pre-existing `..._NonCrashedChildDoesNotPreemptLegacyButtons` was inverted to match the new behaviour).
- [x] Deployed DLL grepped for the new UTF-16 strings — `IsRewindableSlot` (4×), `RewindableSlot` (5×), `not a rewindable slot` (1×), new tooltip (1×).
- [ ] In-game playtest on save `s4` from logs `2026-04-23_2223_unfinished-flights-rewind/`: "Kerbal X Probe" row should now show `Rewind` (not `R`), confirmation dialog should mention the staging split, reload should drop the player into FLIGHT on the probe at UT ≈ 156.
- [ ] In-game: the four early "Kerbal X Debris" rows should remain without a Rewind button (their parent breakups were single-controllable, no RP authored — expected).


vl3c:
Closing per design re-alignment.

After looking at the booster trace more carefully together, the 'Kerbal X Probe' (pid 2969885053) WAS actually destroyed by KSP during the upper-stage time warp — the scene-exit finalizer hit `PatchedConicSnapshot: solver unavailable` (NullSolver), fell back to 'live orbit state' which returned alt=-594km (vessel effectively at Kerbin center = destroyed vessel fingerprint), ran the extrapolator which hit `horizon-cap` and silently defaulted to `terminal=Orbiting`. That's the real bug (B).

The intended scope of 'Unfinished Flights' / the Rewind button is **destruction or loss only** — flights that ended badly (crashed booster, EVA kerbal without a parachute, failed landing). Orbital EVAs, in-orbit dockings, and other splits that end in a stable state do NOT need a Rewind affordance; those recordings spawn their end state and are done. Broadening the predicate to include non-crashed slots (as this PR did) would pollute the list with every routine orbital separation, which is the opposite of what the feature is for.

Reopening as a new branch that:
1. Keeps the narrow `IsUnfinishedFlight` predicate (reverts the broadening in this PR)
2. Fixes the extrapolator / scene-exit finalizer so a NullSolver+sub-surface fallback classifies the vessel as `Destroyed` instead of silently defaulting to `Orbiting`
3. Clarifies design doc §3.1 / §7.31 to make the 'destruction or loss only' vision explicit and call out stable-end splits as an explicit non-example.

PR #513, opened 2026-04-23
Fix #551 tracking station lifecycle parity

vl3c:
Share the map-presence source resolver between Map View and Tracking Station, add state-vector/materialized-vessel parity handling for TS, and cover the shared behavior with headless tests. Validation: dotnet test Source/Parsek.Tests/Parsek.Tests.csproj.


vl3c:
Review fixes applied: renamed `logOperationName` from the method name to verb labels (`map-presence-initial-create`, `map-presence-pending-create`) in `ParsekPlaybackPolicy`, added headless regressions for the three resolver skip branches the shared helper now owns (state-vector threshold, relative-frame, and state-vector-failure normalization when terminal-orbit fallback is off), and dropped `ParsekTrackingStation.cs` from the `#551` todo Files list since the PR does not modify it. Rebased onto `origin/main` after PR #512 merged; no conflicts. Full test suite 8381/8381. Ready to merge.

PR #547, opened 2026-04-25
Fix #582: state-vector ghost map paths honour Relative frame

vl3c:
## Summary

`GhostMapPresence.CreateGhostVesselFromStateVectors` (`Source/Parsek/GhostMapPresence.cs:1979`) and `GhostMapPresence.UpdateGhostOrbitFromStateVectors` (`Source/Parsek/GhostMapPresence.cs:2047`) unconditionally fed `TrajectoryPoint.latitude/longitude/altitude` into `body.GetWorldSurfacePosition`. Those fields are reused as anchor-local XYZ offsets when the originating `TrackSection` is `ReferenceFrame.Relative` (`TrajectoryPoint.cs:13-15`, `TrackSection.cs:34-38`), so a state-vector ghost that traversed a docking/rendezvous segment snapped to a meaningless body-surface lat/lon.

- The flight-scene playback path (`ParsekFlight.InterpolateAndPositionRelative` line 13751) and the diagnostic summary at `GhostPlaybackEngine.cs:3771` already use `TrajectoryMath.ResolveRelativePlaybackPosition` for this. This PR extracts a pure-static helper `GhostMapPresence.ResolveStateVectorWorldPositionPure` that delegates to the same canonical resolver, plus a KSP-dependent wrapper that does the body / anchor lookups.
- Both production methods now log the resolved branch (`absolute` / `relative` / `orbital-checkpoint` / `no-section`) and refuse cleanly when the anchor pid cannot be resolved or when state-vector data was somehow sourced from an `OrbitalCheckpoint` section.
- `UpdateGhostOrbitFromStateVectors` gained an `IPlaybackTrajectory traj` parameter so it can look up the section. Updated call sites: `GhostMapPresence.UpdateTrackingStationGhostLifecycle` (already pre-gated on `IsInRelativeFrame` at `GhostMapPresence.cs:1733`) and `ParsekPlaybackPolicy.CheckPendingMapVessels` line 1019 (no pre-gate, so the latent defect was actually reachable here).

## Latent fix — did NOT fire in captured playtest

In the `logs/2026-04-25_1314_marker-validator-fix` playtest, every state-vector ghost creation attempt was rejected by the threshold gate (`reason=state-vector-threshold`, `no-state-vector-point`). This bug never fired in that capture, so this PR is a latent defect fix.

## User-visible symptom when it would fire

A ghost map vessel transitions through a `Relative`-frame `TrackSection` (Phase 3b docking / rendezvous) while above the state-vector threshold. The ghost icon snaps to roughly the body surface (because `alt` is small) but at a horizontally-meaningless lat/lon — visually identical to "ghost icon goes inside the planet". This is one contributor to the broader #571 symptom family; #571 stays open for the other (recorder-side checkpoint densification + predicted-tail) parts.

## Test plan

- [x] `cd Source/Parsek.Tests && dotnet test --filter StateVectorWorldFrameTests` — 9 new tests pass (Absolute, Relative v6, Relative legacy v5, Relative anchor-not-found, OrbitalCheckpoint, no-section fallback, Absolute-vs-Relative discriminator, branch-tag round-trip).
- [x] `cd Source/Parsek.Tests && dotnet test` — full suite green (8651 / 8651, baseline + my new tests).
- [x] `cd Source/Parsek && dotnet build` — clean, no new warnings.
- [x] Deployed DLL verified to contain new UTF-16 strings (`state-vector-from-orbital-checkpoint`, `anchor-not-found`, `state-vector-orbit-update`).
- [ ] Live playtest: dock two craft, end a recording mid-rendezvous so the recording's terminal section is `Relative`, warp from the tracking station and confirm the ghost map icon stays attached to the anchor instead of snapping to the planet surface.


vl3c:
## Follow-up: GhostMap observability widening (commit `41d9b271`)

Pushed a follow-up that brings the entire ghost-map lifecycle under a single structured-line contract so a future "ghost icon went weird in map mode" report can be reconstructed from the KSP.log alone.

### New logging sink

`GhostMapPresence.BuildGhostMapDecisionLine(GhostMapDecisionFields)` is the single place every decision passes through. Producers fill the field bag via `NewDecisionFields(action)` which seeds NaN sentinels so unset numeric fields disappear from the line.

Standard fields (always present): `action`, `rec`, `idx`, `vessel`, `source`, `branch`, `body`, `scene`.

Optional slots (populated when the source / branch implies them): `worldPos`, `ghostPid`, `segmentBody / segmentUT / sma / ecc / inc / mna / epoch`, `terminalOrbitBody / terminalSma / terminalEcc`, `stateVecAlt / stateVecSpeed`, `anchorPid / anchorPos / localOffset`, `ut`, `reason`.

### Canonical actions

Use these names so existing greps keep working:

- `create-dispatch`, `create-segment-intent`, `create-segment-done`, `create-terminal-orbit-intent`, `create-terminal-orbit-done`, `create-state-vector-intent`, `create-state-vector-done`, `create-state-vector-skip`, `create-state-vector-miss`, `create-chain-intent`, `create-chain-done`
- `update-segment`, `update-state-vector`, `update-state-vector-soi-change`, `update-state-vector-skip`, `update-state-vector-miss`, `update-terminal-orbit-fallback`, `update-chain-segment`
- `destroy`, `destroy-chain`
- `source-resolve`

### Per-tick summary

Both lifecycle drivers (`UpdateTrackingStationGhostLifecycle` and `ParsekPlaybackPolicy.CheckPendingMapVessels`) call `EmitLifecycleSummary(scope, currentUT)` at end-of-tick, which logs `vesselsTracked / created / destroyed / updated / currentUT / scene` once per ~5 s rate-limited window and resets the counters.

### Coordinate audit

Every world-position computation in `GhostMapPresence.cs` now appears in a structured line that names both the input branch and the resolved world position. The Relative branch additionally carries `anchorPid / anchorPos / localOffset`, so the post-hoc reader can verify that anchor-local data was rotated through the correct contract instead of being misread as lat/lon.

### Destroy lines

`RemoveGhostVesselForRecording` and the chain-pid `RemoveGhostVessel` now stash a `LastKnownGhostFrame` on every successful position/update and read it back at destroy time, so the destroy line carries the trailing branch / worldPos / anchorPid / body alongside the existing reason and ghostPid.

### Tests + state

- `Source/Parsek.Tests/GhostMapObservabilityTests.cs` — 32 new tests covering the builder field contract, the resolution-branch translator, the lifecycle-summary helper, the per-branch coordinate contract (Absolute uses surface lookup, Relative uses anchor + canonical helper, distinct world positions for identical input data), and a smoke matrix asserting every canonical action round-trips.
- Full suite: 8683 / 8683 passing.
- Build clean, no new warnings.
- Deployed DLL verified to contain the new action strings.

### Docs

- `CHANGELOG.md` #582 line extended to mention the observability widening.
- `docs/dev/todo-and-known-bugs.md` #582 gets an "Observability:" paragraph documenting the canonical actions, field shape, branch-tag convention, and the `EmitLifecycleSummary` contract so future agents extend rather than duplicate the logging surface.

PR #567, opened 2026-04-25
Relax in-game sidecar-version test for legacy-loop-migrated recordings

vl3c:
## Summary

The in-game test `SaveLoadTests.CurrentFormatTrajectorySidecarsProbeAsBinary` was failing in every scene for any recording that had been migrated from a pre-v4 loop save: a legacy `loopIntervalSeconds` was bumped to v4 (launch-to-launch period) in memory by `RecordingStore.MigrateLegacyLoopIntervalAfterHydration`, but the on-disk `.prec` sidecar was left at v3. The test compared the two values for strict equality, treating two distinct concepts as the same number:

- `probe.FormatVersion` is the **on-disk binary-encoding version** stamped at the last `.prec` write.
- `rec.RecordingFormatVersion` is the **in-memory semantic version**, which post-load migrations can promote without rewriting the sidecar.

The v4 bump was metadata-only (the binary layout is byte-identical to v3, see `TrajectorySidecarBinary.cs:32-39`), so leaving the sidecar at v3 is correct. The production read path in `TrajectorySidecarBinary.Read` already enforces `probe.FormatVersion <= rec.RecordingFormatVersion` via its promote-only branch (`TrajectorySidecarBinary.cs:226-227`), so no production code change was needed — only the test's strict-equality assumption was wrong. A grep for `probe.FormatVersion` confirmed no other production caller makes the same assumption.

The test now asserts the asymmetric contract that actually holds:

    probe.FormatVersion <= rec.RecordingFormatVersion <= RecordingStore.CurrentRecordingFormatVersion

with an inline comment documenting the on-disk-binary vs. in-memory-semantic split. New xUnit regression `TrajectorySidecarProbeVersionContractTests` covers both:

- the post-legacy-loop-migration case where the .prec stays at v3 and the recording is at v4 in memory (the asymmetric contract);
- the freshly-written-at-current-version case where the contract reduces to equality.

## Test plan

- [x] `cd Source/Parsek.Tests && dotnet test` — 8818 passed, 0 failed
- [x] `cd Source/Parsek && dotnet build` — succeeds; deployed DLL verified to contain the new assertion strings (`has on-disk binary version`, `probe should never exceed the recording`, `latest known schema is`) and not the old (`should keep its on-disk format version`)
- [ ] Reload `logs/2026-04-25_2147`'s save, run Ctrl+Shift+T, confirm `SaveLoadTests.CurrentFormatTrajectorySidecarsProbeAsBinary` passes for `1bbb50cf98654a23a60b3248848b0301` ("Learstar A1") in EDITOR / FLIGHT / MAINMENU / SPACECENTER / TRACKSTATION

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ All notable changes to Parsek are documented here.
 
 ### Bug Fixes
 
+- In-game `SaveLoadTests.CurrentFormatTrajectorySidecarsProbeAsBinary` no longer flags legacy-migrated recordings whose binary `.prec` sidecar predates the v4 loop-interval semantic bump. The on-disk binary version is allowed to lag the in-memory recording version because v4 was a metadata-only change.
+
 - `#591` Missed-vessel-switch recovery no longer floods `KSP.log` with redundant recorder-state snapshots. Identical recovery frames now collapse into 5-second `suppressed=N` summaries while normal vessel-switch diagnostics are unchanged.
 
 - `#571` Long on-rails OrbitalCheckpoint warp sections now get derived trajectory samples every 5 degrees of true anomaly, so ghost icons follow the checkpoint window instead of replaying one sparse Kepler segment. The representative 22 ks Kerbin warp adds 42 points and preserves them through format-v6 `.prec` round trips.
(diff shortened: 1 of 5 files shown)


vl3c:
Tightened the contract — the lag exception is now exactly v3 sidecar / v4 recording (the documented metadata-only migration). Any other lag fails the assertion. New xUnit cases pin v5/v6 rejection.

PR #570, opened 2026-04-25
Commit live rewind canary tree so it actually lands before the wait

vl3c:
## Summary

- The in-game `RewindToLaunch_PostRewindFlightLoad_KeepsFutureFundsAndContractsFiltered` canary added in #527 called `flight.StopRecording()` and waited for the recording in `RecordingStore.CommittedRecordings`. `ParsekFlight.StopRecording` only stops the underlying `FlightRecorder`; commit happens via `CommitTreeFlight` (or the scene-exit MergeDialog path), which is the only call that runs `FinalizeTreeRecordings`, copies `RewindSaveFileName` to the tree root, and adds the recording to `CommittedRecordings`. The 10 s wait timed out on its first real execution with `committedBefore=294, committedNow=294, isRecording=False` (2026-04-25_2147 playtest).
- The test was added in `b3d73408` but the runner skipped it in batch mode (`AllowBatchExecution = false`), so the contract mismatch never surfaced until the operator ran it isolated for the first time on 2026-04-25.
- Drive `flight.CommitTreeFlight()` instead of `flight.StopRecording()` so the launch tree finalizes, the rewind save filename reaches the root recording, and `InitiateRewind` has a real rewind owner to resolve. The underlying post-rewind FLIGHT cutoff decision is already pinned by the headless `RewindUtCutoffTests` suite, so no new xUnit pin is required for the fix itself.

## Test plan

- [x] `cd Source/Parsek.Tests && dotnet test` (Passed: 8816, Failed: 0).
- [x] `cd Source/Parsek && dotnet build` (Build succeeded; auto-deploy to KSP `GameData` was locked by a running KSP instance and will need a manual copy before the next playtest).
- [ ] In-game: run `Run All + Isolated` in a Career-mode FLIGHT save with a fresh PRELAUNCH vessel and confirm the canary now reaches the post-rewind FLIGHT cutoff log assertion and passes.


vl3c:
P3 cite-fix landed in 01be1531. Verified line numbers against the actual log files before editing:

- parsek-test-results.txt: FAIL row at line 30 + message at line 31. Updated 25 -> 30-31.
- Player.log: 'Running:' is at line 444977 (line 444980 is the next BgRecorder log), 'FAILED:' is at line 461516 (line 461519 is the 'Restoring batch FLIGHT baseline' log that follows). The existing 444977-461516 range already brackets both lines correctly, so it was left as-is. (Verified with both ripgrep -n and awk NR.)

PR #683, opened 2026-04-30
Batch ghost audio playback cap enforcement

vl3c:
Summary: Batch replayed ghost engine audio power changes before enforcing the per-ghost loop-source cap, avoiding same-frame Play/Stop churn during stage-separation event batches. Keep SetEngineAudio immediate by default for direct callers, update deferred restore/visibility paths to enforce once after batch, and add GhostAudioTests regression coverage plus changelog entry. Validation: dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter GhostAudioTests passed 49 tests; dotnet build Source/Parsek.Tests/Parsek.Tests.csproj --no-restore succeeded; dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName!~InjectAllRecordings --no-build passed 10052 tests. Full unfiltered xUnit was attempted earlier and only SyntheticRecordingTests.InjectAllRecordings failed because the live KSP.log was locked by the running game.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to Parsek are documented here.
 ### Bug Fixes
 
 - Ghost trajectory rendering: Upper-stage ghosts now stay aligned through stage separation during Re-Fly and Watch mode. The trajectory line no longer bounces across the relative-to-absolute transition.
+- Ghost engine audio now batches simultaneous replayed engine events before enforcing the per-ghost loop-source cap, preventing same-frame Play/Stop churn during stage separation from producing audible crackles.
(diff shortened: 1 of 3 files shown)


vl3c:
Addressed review follow-up: tightened the test so deferred SetEngineAudio calls are observed before a single batch cap enforcement, added a headless testing override for the Unity-heavy cap path, and made SetEngineAudio report whether it actually touched a tracked audio key so missing-key/no-audio cases do not schedule wasted batch enforcement. Validation: dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter GhostAudioTests passed 49 tests; dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName!~InjectAllRecordings" --no-build passed 10052 tests.

PR #685, opened 2026-04-30
Fix in-place Re-Fly STASH cleanup

vl3c:
## Summary
- Promote session-created rewind points after in-place Re-Fly marker/state finalization succeeds, immediately before the reaper removes closed STASH slots.
- Add regression coverage for the in-place destroyed-origin plus crash-survivor session RP case, including log assertions and the post-marker-clear promotion ordering.
- Update CHANGELOG and todo/known-bugs docs with the player-facing fix and diagnosis.

## Tests
- dotnet build from Source/Parsek
- dotnet test from Source/Parsek.Tests (10056 passed)
- Verified the deployed KSP DLL hash-matches this worktree build and contains the updated UTF-16 log string.


vl3c:
Agent review summary:
- Initial independent review found one P2 ordering risk: session-created RP promotion ran before in-place marker/state finalization, so an interrupted merge could leave a promoted RP while the Re-Fly marker was still active.
- Addressed in b0c53272 by moving RP promotion after marker clear / force-Immutable / stashed-slot cleanup, immediately before RewindPointReaper.ReapOrphanedRPs(). The regression test now asserts the promotion log occurs after the ReFlySession End reason=merged marker-clear log.
- Follow-up agent review found no issues in the updated diff.

Validation after the follow-up:
- Focused xUnit regression passed.
- Full dotnet test passed: 10056/10056.
- Deployed KSP DLL hash-matches this worktree build and contains the updated UTF-16 log string.

PR #727, opened 2026-05-02
Preserve finalizer orbit tails through Re-Fly merges

vl3c:
Summary: preserve scene-exit finalizer predicted orbit tails when Re-Fly/Watch merges rebuild flat points from track sections, including stale flat Points cases; add a regression test for the collected Kerbal X failure shape; update changelog and known-bugs docs. Validation: dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter SessionMergerTests passed; dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName!~InjectAllRecordings passed. Full unfiltered suite is blocked by running KSP locking KSP.log for InjectAllRecordings.


vl3c (commented):
Review pass: no blocking issues found. The fix is scoped to SessionMerger flat-sync recovery, preserves only a monotonic all-predicted orbit suffix starting after the resolved TrackSection payload, and leaves stale flat Points behind by rebuilding from sections. Regression coverage matches the collected Re-Fly/Watch failure shape. Remaining risk is runtime-only confirmation inside KSP for the exact Watch explosion path; headless SessionMerger coverage and the full non-Injection suite passed.

vl3c:
Review follow-up pushed in 684806c0. Addressed: replaced exact SessionMerger orbit-segment comparison with tolerance-based matching for UT/orbital/quaternion/vector fields; made the section-end tail guard explicit and removed the negative-infinity empty-section footgun; added coverage for checkpoint-prefix predicted tails with serialization-scale drift; added rejection coverage for non-predicted, non-monotonic, and pre-section-start tail shapes; updated CHANGELOG and known-bugs wording. Validation: dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter SessionMergerTests passed 72/72; dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName!~InjectAllRecordings passed 10396/10396.

vl3c (commented):
Follow-up review on 684806c0: no blocking issues found in the review fixes. The tolerance-based prefix comparison is local to SessionMerger's finalizer-tail fallback, the section-end guard now fails closed on empty sections, and the new tests cover the previously untested checkpoint-prefix branch plus unsafe-tail rejection paths. Residual risk remains runtime-only Watch verification on the original save.

PR #708, opened 2026-05-01
Rework ghost anchors around recording chains

vl3c:
## Summary
- Add the v11 `anchorRecordingId` schema/sidecar plumbing and staged recorded-data `RelativeAnchorResolver`.
- Teach foreground/background recorders to choose recording-id anchors, prioritizing sealed immutable recordings, same-lineage/same-vessel ghosts, then nearby ghosts.
- Route non-loop Relative playback and production boundary resolution through recorded anchor chains instead of live vessel PIDs, while keeping loop live-anchor behavior explicitly fenced.

## Validation
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName~TrackSectionSerializationTests|FullyQualifiedName~TrajectorySidecarBinaryTests|FullyQualifiedName~FormatRoundtripTests|FullyQualifiedName~FormatVersionTests|FullyQualifiedName~RecordingBuilderV6Tests|FullyQualifiedName~FlightRecorderFormatVersionTests|FullyQualifiedName~SerializationTrackSectionCheckpointTests|FullyQualifiedName~SessionMergerTests|FullyQualifiedName~RelativeAnchorResolverTests"`
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName~AnchorDetectorTests|FullyQualifiedName~BackgroundTrackSectionTests|FullyQualifiedName~EnvironmentTrackingIntegrationTests|FullyQualifiedName~FlightRecorderStructuralEventTests"`
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName~GhostPlaybackEngineTests|FullyQualifiedName~RelativeAnchorResolverTests|FullyQualifiedName~ProductionAnchorWorldFrameResolverTests"`
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj`

## Runtime follow-up
- KSP-only validation still needed for the v11 watch-separation fixture through UT 41.74-46.74 with `[PlaybackTrace]`, plus Watch/Re-Fly/Map-KSC recorded-anchor checklist items documented in the plan.
- Phase D behavior deletion is intentionally not included; it remains gated on D.0 product-behavior confirmation.

PR #732, opened 2026-05-02
Fix in-place Re-Fly discard restore

vl3c:
Summary
- Captures a full pre-Re-Fly original recording snapshot for in-place continuation sessions and persists it in the tree record codec.
- Restores that snapshot when Discard is chosen from the merge dialog, with the old trim behavior left as fallback for older active sessions.
- Clears orphaned original snapshots on commit/load and preserves them across active recorder refreshes.

Investigation
- Newest PR #708 logs showed Discard ran with active == origin and trimmed/saved the origin sidecar to one point and zero track sections.
- The collected evidence bundle is under logs/2026-05-03_0039_pr708-refly-discard-zero-second.

Validation
- Focused Re-Fly/marker/merge tests passed: 121/121.
- Full suite without SyntheticRecordingTests.InjectAllRecordings passed: 10579/10579.
- Full suite ran 10581 passing and failed only because InjectAllRecordings refused to purge the live KSP save while KSP.log was locked by the running game.


vl3c:
Superseded by #733, which targets main directly.

PR #747, opened 2026-05-04
Fix Watch availability after rewind

vl3c:
## Summary
- keep row, group, and Timeline Watch buttons actionable for future inactive recordings by routing through Fast Forward plus deferred Watch
- target FF plus Watch at the first ghost activation UT, not the recording's wider semantic StartUT, so widened ExplicitStartUT lead-in gaps enter Watch reliably
- clear stale deferred Watch handoffs when a later valid fast-forward supersedes them
- clarify Watch eligibility logs by separating playback state hasGhost from loaded visual object ghostObject
- document the log-package root cause in CHANGELOG and known-bugs notes

## Review Follow-up
- addressed item 1: confirmation logs now report the resolved FF plus Watch target UT
- addressed item 2: each valid fast-forward clears any older pending Watch handoff before arming the current one if needed
- addressed item 3: group aggregate FF plus Watch computes activation UT once per candidate
- confirmed item 4 as intended: group FF plus Watch is an aggregate fallback only when no descendant is directly watchable
- prior GPT-5.5 xhigh narrow re-review found no remaining actionable findings before this additional review pass

## Validation
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter RecordingsTableUITests|TimelineWindowUITests|FastForwardTests|FlightPlaybackExplainabilityTests (189/189 passed)
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter FullyQualifiedName!~InjectAllRecordings (10697/10697 passed)
- dotnet build Source/Parsek.Tests/Parsek.Tests.csproj --no-restore (passed)

## Note
- Full unfiltered dotnet test remains blocked by the running KSP instance locking KSP.log / deployed Parsek files; InjectAllRecordings refuses to purge live save recording sidecars as designed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to Parsek are documented here.
 
 ### Bug Fixes
 
+- Watch buttons remain actionable immediately after rewinding before future ghosts have activated. Row, group, and Timeline `W` buttons now route future inactive recordings through the existing fast-forward confirmation and deferred Watch handoff, targeting the first ghost activation UT so the camera enters Watch after playback has spawned and positioned the target ghost. Superseding fast-forwards clear any older deferred Watch target, and Watch availability logs now distinguish an active playback state from a loaded ghost visual object.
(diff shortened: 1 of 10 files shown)


vl3c:
Closing per request: this approach changes too much by adding a Fast-Forward + Watch path. We should revisit the original bug with a narrower fix focused only on why Watch disappeared after rewind.

PR #762, opened 2026-05-06
Fix Re-Fly tree persistence on pending commits

vl3c:
## Summary

- intercept finalized pending trees before stock scene transitions so the merge/discard dialog appears in Flight instead of after Space Center / Tracking Station loads
- update same-ID committed trees only for validated superset/topology updates, preserving exact/equivalent/poorer duplicate no-ops
- add regressions for Re-Fly tree persistence, duplicate/stale same-ID tree guards, and live pending-tree scene-exit decisions

## Log Finding

- collected `logs/2026-05-06_1938_merge-dialog-next-scene` and filtered out `[GhostVisual]` / `[GhostRenderTrace]` entries into `KSP.filtered-render-trace.log` and `Player.filtered-render-trace.log`
- confirmed PR #750's live active-tree intercept worked later in the same session, but the first destruction path had already stashed `PendingTreeState.Finalized` and cleared `activeTree`
- confirmed the Re-Fly tree persistence issue matches PR #751's fork model exposed through the older #401 `ReconciliationBundle` restore of a stale committed tree copy

## Tests

- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName~TreeCommitTests|FullyQualifiedName~SceneExitInterceptorTests|FullyQualifiedName~CommitFlowTests.CommitTree_DuplicateTree_MergerNotRunTwice"`: 113 passed
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName!~SyntheticRecordingTests.InjectAllRecordings"`: 10839 passed

## Review

- GPT-5.5 xhigh agent review found one P1 in the initial same-ID replacement guard
- fixed by adding validated superset/topology replacement logic plus equivalent-copy and poorer/stale-copy regressions
- follow-up GPT-5.5 xhigh review found no remaining blockers

## Notes

- full unfiltered xUnit was not runnable while KSP is open because `SyntheticRecordingTests.InjectAllRecordings` needs to purge a locked `KSP.log`
- builds/tests emitted expected post-build copy warnings for locked deployed KSP files (`Parsek.dll`, `Parsek.pdb`, `Parsek.version`); the test assemblies still passed


vl3c:
Thanks for calling these out. The optimizer-split findings are valid, but they are intentionally out of scope for #762. This PR is limited to the finalized-pending-tree pre-transition dialog and same-id committed-tree topology persistence guard.

The split/rollback issue is fixed in the stacked follow-up PR: https://github.com/vl3c/Parsek/pull/763. That PR is based on this branch (ix-refly-tree-persistence) so reviewers can look only at the optimizer split deferral and rollback sweep without re-reviewing #762's dialog/tree-persistence changes.

I’d prefer to keep #762 scoped as-is and merge it first, then review/merge #763 on top.

PR #767, opened 2026-05-07
PR 2/5: harness debris baselines + same-chain continuation

vl3c:
## Summary

Second slice of the recording & ghost policies refactor — see `docs/dev/plans/recording-and-ghost-policies-refactor-plan.md`. Adds four scenarios to the resolver-level harness from PR 1. No behavior changes.

> Stacked on top of PR #766 (PR 1 harness skeleton). Base is `debris-rendering-pr1-harness`.

Three of the four scenarios capture today's BROKEN debris composition. After PR 3b lands the parent-anchor contract, the scenarios will be rewritten to anchor debris to its parent and set `DebrisParentRecordingId`; their SHA-256 baselines will change and the diff will reset them with a justification comment.

- **Scenario 4** — focused-vessel debris with wrong anchor (plan §"Bug evidence" failure mode #1). `ParsekFlight.CreateBreakupChildRecording` today sets `anchorRecordingId` to "nearest eligible vessel at sample time"; when that's not the parent, the Relative offsets encode in the wrong frame. Encodes today: anchor is `other-vessel` rather than `focused-parent`, world position composes to `(301 + ut, 50, 0)`.
- **Scenario 5** — background-vessel debris with wrong anchor. The `BackgroundRecorder` counterpart of scenario 4. Same bug, different creation site. Different positions/anchors so a bug that affects only one creation site lights up only one baseline.
- **Scenario 7** — Re-Fly debris with wrong anchor (plan §"Bug evidence" failure mode #2). `RestoreActiveTreeFromPending` removes the active Re-Fly recording from `tree.BackgroundMap`, so new debris shed mid-Re-Fly anchors to whatever pre-Re-Fly ghost is loaded. Encodes today: `marker.ActiveReFlyRecordingId` is set, but the debris's anchor is `pre-refly-ghost` (not the active recording), so `TryResolveActiveReFlyAnchorRecording` does NOT fire walk-back — resolver composes against the wrong pre-Re-Fly ghost, producing `(501 + ut, 200, 0)`.
- **Scenario 10** — same-chain continuation. Non-debris invariant. `first-half` (UT [0,10]) and `second-half` (UT [10,20]) on the same `ChainId`; child Relative is anchored to `first-half` but sampled over UT [10,20]. The resolver walks past first-half's range into second-half via `TryResolveSameChainContinuationPose`. **Must NOT change in PR 3b** (per plan's "scenarios 1, 2, 3, 6, 8, 9, 10 unchanged"). Includes a sanity log assertion for `Anchor recording continued through same-chain successor`.

## Test plan

- [x] `dotnet test Source/Parsek.Tests` — all 10,744 tests pass; 4 new tests added on top of PR 1's 4.
- [x] Verified scenario 7 composes against the wrong pre-Re-Fly ghost (501..511, NOT 101..111 which would be the correct walk-back composition).
- [x] Verified scenario 10's resolver emits the same-chain continuation log line for every sampled UT.
- [x] Hashes are deterministic across runs.


vl3c:
Closing: head commit d82c9476 is already on origin/main (landed transitively via the PR3a/3b/3c merges). No GitHub merge needed.

PR #780, opened 2026-05-09
Investigate radial debris origin alignment

vl3c:
## Summary
- document retained-log evidence and a diagnostics-first plan for radial side-booster debris origin alignment
- fix GhostAppearance recording-start diagnostics so flat fallback points inside Relative sections are not logged as Relative offsets
- add bounded joint-break and decouple-created diagnostics for root/child PID, attach-node, and seed-source correlation

## Validation
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter GhostVisualFrameTests
- dotnet test Source/Parsek.Tests/Parsek.Tests.csproj --filter "FullyQualifiedName!~InjectAllRecordings"

Full unfiltered dotnet test was attempted; InjectAllRecordings failed because KSP.log is locked by the running game. I did not close or restart KSP.

PR #784, opened 2026-05-09
Guard relative playback against fast anchor rotation

vl3c:
## Summary
- Active `RecordedRelative` playback now detects abnormal anchor-local offset speed between adjacent recorded frames and routes that segment through the section's v7 `absoluteFrames` shadow instead of lerping in anchor-local Cartesian coordinates.
- Targets the radial-booster fast-parent-rotation case where the recorded parent anchor resolves correctly but linear interpolation between endpoint-local offsets collapses the world-space separation through the midpoint.
- Carved out of #780 because it touches generic Relative playback (shared with Re-Fly rendering), which is broader than #780's recorder-side radial-debris fixes.

## Invariants preserved
- Anchor resolution stays the gatekeeper. The new guard runs **after** `TryResolveRecordedRelativeAnchorPose` succeeds at `ParsekFlight.cs:21657`. If anchor resolve fails, the existing `TryUseRelativeAbsoluteShadowFallback` retires v12+ parent-anchored debris exactly as before.
- The guard calls `TryInterpolateRelativeAbsoluteShadow` **directly**, not the retire-bearing fallback helper. `ShouldRetireOnRecordedParentAnchorMiss` is identity-based and would retire on any v12+ debris frame — it must not be the path used when the anchor itself is fine and we just want better interpolation.
- Threshold: `RecordedRelativeFastAnchorLocalMotionGuardMetersPerSecond = 100.0`. In the radial-debris repro the offending offsets imply ~1006 m/s, an order of magnitude clear of normal debris-vs-parent rates (10-70 m/s).
- v6 recordings without `absoluteFrames` keep the existing lerp path (no shadow available).
- Per-frame `GhostPosEntry.mode` reflects the data path used (`PointInterp` for shadow-routed frames). Future frames re-evaluate anchor resolution before the guard can fire again, so the v12+ retire contract is not weakened.

## Validation
- `dotnet build Source/Parsek/Parsek.csproj` — clean
- `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj` — 10992 passed, 0 failed
- New unit coverage in `RelativeAnchorResolutionTests` for the predicate + log-assertion for the guard's verbose line
- **Re-Fly regression check still owed**: this changes a code path shared with Re-Fly playback. Before merge, an in-game smoke specifically targeting active Re-Fly Relative-frame playback (especially under fast anchor rotation, e.g. ascent-stage attitude maneuvers) is needed to confirm no regression in the Phase D direction.

## Test plan
- [ ] Radial booster decouple + ascent attitude maneuver: verify debris no longer oscillates during fast-rotation segments (the original symptom).
- [ ] Active in-place Re-Fly with non-debris parent-chain ghost: verify the ghost still renders at recorded coordinates (Phase D D.0 invariant).
- [ ] Loop Relative playback (live anchor): verify guard does not fire (no `absoluteFrames` engagement on loop sections).
- [ ] v6 recording with `absoluteFrames=null`: verify legacy lerp path still used.


vl3c:
Closing as superseded. The useful part of this PR was the idea of routing unsafe parent-relative playback through recorded absolute-shadow frames, and that landed in the later parent-anchored debris stack instead: #800 added shadow routing for tumble windows and #803 made v12+ parent-anchored debris render from covering absoluteFrames unconditionally while keeping the tumbling-parent gate only for FX suppression. That narrower path avoids carrying this generic RecordedRelative fallback into Re-Fly / non-debris systems that already work.

PR #812, opened 2026-05-10
Phase 1: activation-flow observability for controlled-ghost slide investigation

vl3c:
## Summary

Phase 1 of the [`fix-controlled-ghost-init-slide`](docs/dev/plans/fix-controlled-ghost-init-slide.md) plan. Permanent observability investment in `GhostRenderTrace` so the user-reported watch-mode controlled-vessel ghost slide (`logs/2026-05-10_1713`, Kerbal X Probe ghost #8) can be empirically pinned to one mechanism before Phase 3 ships any production fix.

Closes the four observability gaps in the existing tracer:
1. Activation hide/activate decision was logged via standalone `ParsekLog.VerboseRateLimited`, not aligned with the structured trace by frame number — new `EmitActivationDecision` phase emit.
2. `AfterUpdate` exposed the visible (clamp-resolved) `playbackUT` only, with no field for raw UT, lead, or whether the clamp fired — three new appended fields.
3. No detailed window opened on the hidden→visible transition — new `activation-transition` window opens at first-visible for 1.0 s.
4. `dM` mixed hidden→hidden, hidden→visible, visible→visible deltas — per-state `lastHiddenPosition` + `hiddenPoseDelta` field on the activation transition row.

No production behaviour change. Tracer is gated behind the existing `Settings → Diagnostics → Ghost render tracing` toggle.

## Changes

**`Source/Parsek/GhostRenderTrace.cs`**
- New `EmitActivationDecision(...)` structured phase emit. Carries `rawPlaybackUT`, `visiblePlaybackUT`, `activationStart`, `activationLead`, `visibleLead`, `clampFired`, `hidden`, `hideReason`, `framesRemaining`, `transition` (`hidden` / `visible` / `first-visible`), `prevHiddenPos`, `hiddenPoseDelta`. The `first-visible` transition is logged at `Info` (rest at `Verbose`) and opens an `activation-transition` detailed window of `ActivationTransitionWindowSeconds = 1.0`.
- New `TraceState` fields: `hasLastHiddenPosition`, `lastHiddenPosition`, `lastHiddenPositionFrame`, `wasHiddenLastActivationDecision`. Per-(recordingId, ghostIndex) granularity via the existing `BuildStateKey`.
- `EmitPostUpdate` signature gains optional trailing parameters `double rawPlaybackUT = double.NaN` and `double activationStartUT = double.NaN`. Three new fields appended to the line: `rawPlaybackUT`, `visibleLead`, `clampFired`. Default NaN treats raw == visible and `clampFired = false` so unmigrated callers stay compile-safe.
- New `IsDetailedWindowOpenForTesting` and `FrameCounterOverrideForTesting` test seams.
- New ECall isolation helpers (`UnityFrameCount`, `ReadGhostPosition`, `ReadGhostRotation`, `IsGhostActiveSelf`) wrap Unity-native references so xUnit JIT verification of the trace emitters passes. Production paths unchanged — these helpers are only invoked when `state.ghost` is non-null.

**`Source/Parsek/GhostPlaybackEngine.cs`**
- `RenderInRangeGhost` (non-loop) emits `EmitActivationDecision` immediately before the `if (initialActivationHidden)` branch. The retired short-circuit at `:1231-1236` (which runs `ApplyFrameVisuals(skipPartEvents:true, suppressVisualFx:true)` and exits) naturally gates the emit out of retired frames per plan §1a.
- `SynchronizeLoadedGhostForWatch` captures the previously-discarded hide reason via a named local instead of `out string _`, and emits `EmitActivationDecision` with `rawPlaybackUT == visiblePlaybackUT == playbackUT` (this path doesn't call `ResolveVisiblePlaybackUT`, so `clampFired` is invariant `false`). Watch-sync rows are activation-only (no surrounding `FrameStart` / `AfterUpdate`) — accepted v1 asymmetry per plan §1a.
- All four `EmitPostUpdate` call sites (`RenderInRangeGhost` non-loop + three loop paths) updated with named-argument `rawPlaybackUT` and `activationStartUT` pass-through. Loop sites pass `loopUT` for both because `ResolveVisiblePlaybackUT` runs only at `:999`, `:1138`, `:5094` (verified by source walk).
- New internal `SynchronizeLoadedGhostForWatchForTesting` shim exposes the otherwise-private watch-resume entry point for in-game test coverage.

**`Source/Parsek.Tests/GhostRenderTraceTests.cs`**
- Six new tests pinning the Phase 1 contracts:
  - `EmitActivationDecision_HiddenFrame_EmitsExpectedFields` — full field-shape assertion for a hidden emit.
  - `EmitActivationDecision_FirstVisibleTransition_EmitsHiddenPoseDelta` — drives two hidden frames at distinct positions then a visible frame; asserts `transition=first-visible` and `hiddenPoseDelta` from the last hidden pose.
  - `EmitActivationDecision_OpensDetailedWindowOnTransition` — asserts the activation-transition window opens at the first-visible transition, covers the next 1.0 s, AND closes after.
  - `EmitPostUpdate_AppendsRawPlaybackUTAndLeadAndClamped` — clamp-firing case with `rawPlaybackUT != playbackUT`.
  - `EmitPostUpdate_DefaultRawPlaybackUT_ClampFiredFalse` — pins the default NaN contract.
  - `EmitPostUpdate_LoopPathInvariant_ClampFiredFalse` — fan-out test for the three loop call sites; locks the loop-path invariant at the trace API level so a future refactor adding `ResolveVisiblePlaybackUT` to a loop path fails this test.

## Out of scope (Phase 2 / 3)

- Phase 2: investigation step using the enhanced trace. Replay `s14` save through the probe-decouple → watch-frame-80871 sequence with diagnostics on, walk the resulting `phase=ActivationDecision` lines through the plan's decision matrix, pick the Phase 3 fix shape.
- Phase 3: actual fix (leading candidate from rev. 2 is `InitialActivationHiddenMinimumFrames = 2 → 3` in `ParsekConfig.cs`). Shape contingent on Phase 2's findings.
- Engine-level watch-sync test in xUnit: `SynchronizeLoadedGhostForWatch`'s `HasLoadedGhostVisuals(state)` guard requires `state.ghost` to be a real Unity `GameObject`; xUnit cannot construct one. Coverage falls to the in-game test runner.

## Test plan

- [x] `dotnet test`: 11304 / 11304 pass.
- [x] Build verified, DLL deployed to `Kerbal Space Program/GameData/Parsek/Plugins/`.
- [ ] In-game smoke pass: `Settings → Diagnostics → Ghost render tracing` ON, replay `s14` save through UT 456.555, capture log bundle, confirm `phase=ActivationDecision` lines appear for the probe ghost across the activation window with the new fields populated.
- [ ] Phase 2 investigation against the captured log bundle.


vl3c:
## Opus review pass + rebase-onto-main

Clean-context Opus review surfaced one P1 (the branch was based on `138cc106` but `origin/main` had moved to `77e41a3b` after PR #808 merged — the diff was reverting PR #808 by accident) plus three P2/P3s on the test and comment quality.

**Resolved:**
- **P1 (blocking):** Merged `origin/main` into the branch via a non-fast-forward merge (commit `1c0b9ef1`). PR #808 is now preserved; the diff against `main` is just the Phase 1 changes.
- **P2 (test rigour):** Tightened `EmitActivationDecision_OpensDetailedWindowOnTransition` so it no longer relies on `EmitPostUpdate`'s `first-seen` gate (which would have made the closed-before assertion satisfied trivially). The test now drives only `EmitActivationDecision`, asserts the window stays closed through the hidden frame, opens at exactly the right UTs after the first-visible transition, and adds a `ShouldEmitPhase` round-trip pinning the 1.0 s ungate window.
- **P2 (comment fix):** Comment in `EmitActivationDecision` claimed the window-first ordering existed "so `ShouldEmitPhase` consulted for this very emit also sees the window open" — but the call uses `force: true`, bypassing the window check. Comment now correctly attributes the ordering to other emitters on the same `currentUT`.
- **P3 (carve-out test):** Added `EmitActivationDecision_RetiredFrame_DoesNotFire` pinning plan §1a's "does NOT fire on retired frames" contract at the API level (the carve-out is structurally enforced by the engine wrapping the call inside the `!retired` else branch).
- **P2 (CHANGELOG note):** Flagged the temporary log-line duplication: the existing `[Engine] initial activation hidden` `VerboseRateLimited` line continues to fire alongside the new structured emit during the Phase 2 investigation cycle. Intentional; a follow-up removes the older line once the structured emit is field-validated.

**Deferred (P3 in review):**
- `EmitPostUpdate_LoopPathInvariant_ClampFiredFalse` exercises the trace API surface, not the engine call sites. Reviewer noted this is acceptable for a tracer-API invariant but doesn't catch engine-side regressions. Engine-driven coverage falls to the in-game test runner pass that's already on the test plan.
- `Vector3.zero` vs `default(Vector3)` inconsistency (engine code uses `Vector3.zero`; trace internals deliberately use `default(Vector3)` for xUnit JIT verification). Cosmetic — production engine code is not subject to JIT verification. Leaving as-is.

**Test results:** 11327 / 11327 pass (count includes new tests from PR #808 plus the new retired-carve-out case).

Branch is fast-forwarded from the previous `f0da8495` tip. Six commits total: 4 plan revisions + Phase 1 implementation + merge + review followup.

vl3c:
## Re-review pass

Clean-context Opus re-review of the post-fix state. Verdict: one P2 (the new retired-carve-out test was a tautology) plus two acknowledged-but-not-blocking P3s (window test moved from emit→log path to predicate-only; comment-fix wording is correct but could name a concrete LateUpdate site). No P1 blockers.

**Fixed:**
- `EmitActivationDecision_RetiredFrame_DoesNotFire` — deleted in `1995de28`. The reviewer was right: the test snapshotted `FindLines` count twice with no call between them and asserted equality, then asserted a recording id appears in zero lines despite never being passed to any emitter. Tautology. A meaningful test would need to drive `RenderInRangeGhost` directly with a real Unity `GameObject` for `state.ghost` — same xUnit limitation that already pushed watch-sync engine coverage to the in-game runner. Replaced with a comment in the same location explaining the gap and the structural guarantee at `GhostPlaybackEngine.cs:1233-1276`.

**Acknowledged, not blocking:**
- Window test now drives only `IsDetailedWindowOpenForTesting` / `ShouldEmitPhase` instead of the full emit→log path. Reviewer noted "the predicate IS the gate" so coverage is substantively correct. A one-line addition (drive `EmitPostUpdate` post-window, assert an `AfterUpdate` line lands) would close the loop end-to-end. Held for a follow-up if a future regression sneaks past the predicate-only assertion.
- Comment-fix re-attribution to "OTHER emitters on the same currentUT (e.g. an immediately-following LateUpdate pass)" is correct — `ParsekFlight.LateUpdate` runs after the engine's update and its `EmitPhase` calls do consult the open window. Reviewer suggested naming a specific call site for concreteness; held.

**Confirmed clean by reviewer:**
- Merge resolution preserved PR #808's Done + Open entries plus the new controlled-ghost slide entry; no duplicates.
- CHANGELOG entries are intact at the right positions.
- Scope discipline holds: `git diff origin/main...HEAD --stat` is exactly the planned six files; engine diff is purely observability emits + one test seam.
- No merge regressions: PR #808's three new files merged cleanly with zero overlap with Phase 1's edits.

**Test results:** trace tests 12 / 12 pass post-fix.

PR is now `1995de28`.

vl3c:
## Re-review #2 follow-up

Two more P2s caught both correctly. Fixed in `db661925`.

**P2 #1 — `EmitActivationDecision` log flood.** Reviewer caught that the engine call site is unconditional inside the `!retired` else branch and `EmitActivationDecision` uses `force: true`, so once a ghost transitioned past first-visible, every subsequent frame emitted a `transition=visible` row for the rest of its lifetime whenever ghost render tracing was on. Fix: early-return in `EmitActivationDecision` when `transition == "visible"` (steady state). Hidden frames + first-visible transition still emit; the steady-state noise is dropped. State updates skipped too — `wasHiddenLastActivationDecision` was already correctly set to `false` on the first-visible emit, so a future re-hide (loop wrap, scene reload) is observed correctly without further bookkeeping.

**P2 #2 — watch-sync `currentUT` time-coordinate mismatch.** Reviewer caught that for looped watch targets, the `playbackUT` parameter is loop-mapped recording time, not real Planetarium UT — so passing it as `currentUT` to `EmitActivationDecision` opened the activation-transition detailed window in the wrong time coordinate and would have missed subsequent `AfterUpdate` / `LateUpdate` emits keying off Planetarium UT. Fix: thread the actual current UT through the watch path. `EnsureGhostVisualsLoadedForWatch` and `SynchronizeLoadedGhostForWatch` both gain a `double currentUT` parameter; both `WatchModeController` call sites pass `Planetarium.GetUniversalTime()`; the watch-sync `EmitActivationDecision` call now passes the real `currentUT` separately from `playbackUT`. Test shim + one existing test updated for the new signature.

**Test results:** 11326 / 11326 pass.

PR is now `db661925`.

vl3c:
## Re-review #3 follow-up

Two more P3s, both fixed in `a8a084e6`.

**P3 #1 — guard call site missed in P2 #2 watch-sync UT threading.** `EnsureGhostVisualsLoadedForWatch` correctly threads `currentUT` through to `EmitActivationDecision` now, but the early-return `TryHandleParentAnchoredDebrisCoverageRetired` guard at the top of the method still passed `playbackUT` for BOTH the `playbackUT` AND `currentUT` params. For looped watch targets the guard's `EmitGuardSkip` log row would carry loop-mapped recording time as `currentUT`, mismatched with the rest of the watch trace. Fix: pass real `currentUT` through to the guard too.

**P3 #2 — CHANGELOG / plan stale on `transition=visible` suppression.** The prior P2 flood fix made `EmitActivationDecision` early-return when `transition == "visible"` (steady state) to avoid a per-frame log flood. CHANGELOG and plan still listed the row vocabulary as `hidden | visible | first-visible` and described `EmitActivationDecision` as covering the full activation lifecycle. Updated both: vocabulary is now `hidden | first-visible` only; CHANGELOG explicitly notes Phase 2 investigators should NOT expect a row per frame after activation completes.

**Test results:** 11326 / 11326 pass.

PR is now `a8a084e6`.

vl3c:
## Final review-pass P3s addressed

All three nice-to-haves from the clean review pass are now fixed in `f07500d8`.

**P3-1 (hoist `Planetarium.GetUniversalTime()`):** Both `WatchModeController` call sites (`:1741` and `:3422`) had two getter calls per watch entry — once for `ResolveWatchPlaybackUT`, once for the new `currentUT` param. Both hoisted to locals.

**P3-3 (tighten window test):** `EmitActivationDecision_OpensDetailedWindowOnTransition` previously asserted `ShouldEmitPhase` round-trip results that were redundant with the `IsDetailedWindowOpenForTesting` assertions two lines up — both consult the same predicate. Rewrote to genuinely pin end-to-end:
- Pre-prime per-state `firstSeenUT` at `currentUT=100` via a primer `EmitPostUpdate`, which isolates the activation-transition window from `initial-window`'s 4 s gating effect at later UTs.
- Drive an actual `EmitPostUpdate` at `currentUT=201.0` (firmly past `initial-window`, inside `activation-transition`) and assert an `AfterUpdate` line lands.
- Drive an `EmitPostUpdate` at `currentUT=500` (past both windows) and assert nothing lands.

The load-bearing claim is now exercised, not just hand-waved by the comment.

**P3-2 (`Planetarium.GetUniversalTime()` always fires):** Plan rev. log gains a note documenting that the threaded getter calls fire on every watch-load entry regardless of `ghostRenderTracing` setting. Below the noise floor, but documents the place to add a tracer-enable check at the call site if a future refactor wants to skip the getter outright when tracing is off. Plan bumped to rev. 7.

**Test results:** trace tests 12 / 12 pass, full suite 11326 / 11326 pass.

PR is now `f07500d8` (10 commits total).

PR #816, opened 2026-05-10
Hide cobubble peer ghost during post-load settle

vl3c:
## Summary

Phase 3 of the controlled-vessel ghost initial-slide fix. Phase 1 ([#812](https://github.com/vl3c/Parsek/pull/812)) shipped permanent observability in `GhostRenderTrace` (the `EmitActivationDecision` structured emit + `AfterUpdate` fields + `hiddenPoseDelta`). Phase 2 captured `logs/2026-05-10_2123/` and ruled out the `ResolveVisiblePlaybackUT` clamp as the cause (`clampFired=false` in every observed slide event).

The smoking gun in that bundle:

```
21:18:09.640 [ReFlySettle] FloatingOrigin.setOffset
  refPos=(-465.223,0.191,-241.438) magnitude=524.14 frame=49808
21:18:09.641 [GhostRenderTrace] phase=AfterUpdate ghostIndex=1 frame=49808
  reason=large-delta dM=489.22 expectedDM=34.92 clampFired=false
21:18:09.641 [GhostRenderTrace] phase=ActivationDecision ghostIndex=1 frame=49808
  transition=first-visible activationLead=0.300 hiddenPoseDelta=489.225
```

The peer ghost was in an active co-bubble blend window (entered at separation UT 135.303), and its activation hide released after two frames — exactly when a Krakensbane-driven `FloatingOrigin.setOffset` shifted the world frame by ~524 m. The peer's first visible frame thus landed on a stale-then-fresh transform pair.

## Fix

Option 1 from the Phase 2 write-up: extend the activation hide to cover the active co-bubble blend window. New `ShouldHoldInitialCoBubbleBlendActivationHidden` predicate slots into the existing `ShouldHoldInitialActivationHiddenThisFrame` precedence chain between `absolute-primer-to-relative` and `activation-settle`. Fires when:

- `CanEvaluateInitialActivationHidden` is true (deferred visibility, `appearanceCount==0`)
- An active co-bubble blend window covers `playbackUT` (resolved via `Parsek.Rendering.SectionAnnotationStore.TryGetCoBubbleTraces`)
- `playbackUT - activationStartUT <= GhostPlayback.InitialCoBubbleBlendActivationHiddenSeconds` (new constant, **0.5 s**)

Reports reason `cobubble-blend-settle`. The two-frame `InitialActivationHiddenMinimumFrames` minimum still applies on top, so a typical hold runs the cobubble window then drops into `minimum-frames` for the residual frames before release. `relative-start`, `absolute-seed-bridge`, and `absolute-primer-to-relative` still win when their predicates fire on the same frame (precedence chain invariant).

Test seam `GhostPlaybackEngine.CoBubbleBlendStartUTResolverForTesting` lets xUnit drive the predicate without standing up the full annotation pipeline; production walks `SectionAnnotationStore.TryGetCoBubbleTraces`. The `[Engine] initial activation hidden` `VerboseRateLimited` line gains a `coBubbleBlend=` field alongside the existing `relativeWindow=` / `absoluteBridgeMax=` / `debrisSeedBridgeMax=` / `minFrames=` constants.

## Plan reference

`docs/dev/plans/fix-controlled-ghost-init-slide.md` — Phase 3 fix landing.

## Test plan

- [x] `dotnet test` — 11418 passed, 0 failed
- [x] New `GhostPlaybackEngineTests` predicate truth-table cases (within hold, past hold, no active blend, resumed-ghost no-op)
- [x] New `GhostPlaybackEngineTests` end-to-end cases (cobubble-blend-settle fires inside window then drops to minimum-frames; relative-start still wins precedence)
- [ ] In-game playthrough: replay the `s14`/`s15` save through the probe-decouple → re-fly load sequence and confirm no visible peer ghost slide on first activation
- [ ] In-game playthrough: confirm no regression on Watch entry for non-cobubble peer ghosts (debris, single-vessel scenes) — should retain existing two-frame minimum hold


vl3c:
Closing — investigation in the follow-up worktree showed the captured `dM=489` event was largely a trace measurement artifact of FloatingOrigin shifts (camera shifts in unison via `OnFloatingOriginShift`, scene-relative motion was `expectedDM` 35 m). The originally-reported s14 slide was on a non-co-bubble ghost so the predicate wouldn't have fired even if the slide were real, and the actual visible-slide source remains unidentified.

The fix was defensive only and doesn't demonstrably address any visible bug — keeping it would make the area noisier and falsely signal "already handled" to the next investigator. The real wins from this arc are PR #812's tracer + the lesson that `dM` measures world-frame motion (which includes FloatingOrigin shifts) rather than scene-relative motion.

PR #823, opened 2026-05-10
Plan debris relative validation

vl3c:
Summary: Add the reviewed current-main debris relative validation plan. It separates fresh post-PR-817 recorder validation from legacy replay validation and avoids treating old sidecars as current-main recorder failures. Review: GPT-5.5 xhigh approved after revision. Validation: docs-only; git diff checks passed before commit.

### DIFF
--- a/docs/dev/plans/validate-debris-relative-current.md
+++ b/docs/dev/plans/validate-debris-relative-current.md
@@ -1,0 +1,91 @@
+# Validate Debris Relative Playback on Current Main
+
+## Scope
+
+Worker D validation pass for the retained debris relative playback reports around stale tails and sparse Relative sections. This plan intentionally does not propose production or test edits unless current-main evidence confirms an open bug.
+
+## Current-main status
+
+- Worktree base: `origin/main` at `c84010d8` (`Merge pull request #818 from vl3c/fix-tail-orbit-state-vector-frame`).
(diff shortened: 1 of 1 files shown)


vl3c:
Closing as stale. PR 824 (`Implement v13 debris frame contract`) replaced the pre-v13 debris-relative contract this validation plan was scoped against:

- `bodyFixedFrames` is now the primary parent-anchored debris render surface (anchor-local `frames` retained for loop-anchored chain replay / fallback).
- Format bumped to v13; v11/v12 sidecars are no longer accepted.
- Old legacy debris shadow gate, tumbling-parent reliability router, persistent coverage-retired guard, and debris Re-Fly settle suppression paths were removed.
- New parent-proximity sampling tiers (`<=250 m` full, `<=500 m` half/entry, `>550 m` Relative exit) and forced Relative-entry sampling with body-fixed primary coverage.

The recorder-persistability invariant the plan called out for further validation ("a parent-anchored Relative section must not outlive recorder-persistable authored coverage") was codified in CLAUDE.md and is now guarded by tests under the v13 contract. PR 832 (rolled in through PR 824's merge chain) additionally fixed the FG recorder LLA staleness that was muddying separation evidence in retained pre-#817 logs.

If a current-main repro turns up a real recorder/playback gap, it'll need to be re-scoped against the v13 contract rather than the v12 surface this plan targeted.

TODO entry for the related controlled-vessel slide flipped to Done in #833.

PR #831, opened 2026-05-12
Skip activation-settle hide for re-fly controlled ghosts with deterministic start

vl3c:
## Summary

Fixes a regression introduced by the May-3 generalization of the activation-settle hide (commit 4e949cea, *Hide fresh watch activation settle frames*). After that change, the previous attempt's controlled ghost would appear `v_recorded × 0.04 s` past its recorded activation-start position when it first became visible at re-fly load — ~80 m at orbital ascent velocities. User-visible: the upper-stage capsule ghost showed up *behind* the live re-fly probe along the recorded velocity vector, instead of at the separation point.

## Mechanism

The May-3 change added a generic `activation-settle` clause that primes the 2-physics-tick fresh-activation hide for every fresh non-debris activation, not just the four UT-window cases (relative-start, absolute-seed-bridge, debris-seed-bridge, absolute-to-relative-primer). For a controlled ghost whose activation section at `activationStartUT` is `ReferenceFrame.Absolute` (or whose flat `Recording.Points` cover the activation UT), the first frame's recorded position is fully deterministic — there is no anchor-resolution race for the settle frames to mask. While the ghost was hidden, `playbackUT` advanced 0.04 s, and the first-visible frame rendered at `recorded_pos(activationStartUT + 0.04 s)`.

Repro in `logs/2026-05-13_0038_refly-probebooster-ghost-behind`, frames 34696–34698:

```
frame 34696  HIDDEN reason=activation-settle  pos=(-236.98,-6.84,-177.80)
frame 34697  HIDDEN reason=minimum-frames     pos=(-303.20,-8.76,-227.39)  dM=82.76 expectedDM=82.76
frame 34698  FIRST-VISIBLE                    pos=(-303.20,-8.76,-227.39)
```

velocity = (-1642, -48, -1258) m/s, magnitude 2068 m/s. **2068 × 0.04 s = 82.76 m**.

## Fix

`GhostPlaybackEngine.IsActiveReFlyControlledTrajectoryWithDeterministicActivationStart` skips the `activation-settle` clause when:

1. A host-supplied probe (`ActiveReFlySessionInProgressProbe`, wired by `ParsekScenario.OnAwake` to `ActiveReFlySessionMarker != null`) reports an active re-fly session.
2. The trajectory is not debris.
3. The active section at `activationStartUT` is `ReferenceFrame.Absolute`, **or** flat `Recording.Points` cover `activationStartUT`.

Only the generic `activation-settle` clause is suppressed. The four UT-window hides above it (`relative-start`, `absolute-seed-bridge`, `debris-seed-bridge`, `absolute-to-relative-primer`) and the debris paths are unchanged. Outside an active re-fly the predicate returns false and the existing `activation-settle` behavior is intact (loop restarts, watch-mode anchoring still take the hide).

The probe is a static `Func<bool>` delegate so the engine keeps its standalone-mod boundary — it never sees the `ReFlySessionMarker` shape, only its presence/absence. Standalone consumers leave the probe null and the carve-out never fires.

## Test plan

- [x] `GhostPlaybackEngineTests` adds five tests pinning the carve-out (`ActiveReFlyControlledAbsolute_SkipsActivationSettle`, `ActiveReFlyControlledFlatPoints_SkipsActivationSettle`, `ActiveReFlyDebris_StillTakesHide`, `ActiveReFlyRelativeSection_StillTakesUtWindowHide`, `NoReFlySession_StillTakesActivationSettleHide`) plus `IsActiveReFlyControlledTrajectoryWithDeterministicActivationStart_PredicateBranches` covering null / debris / null-probe / false-probe / throwing-probe / probe-true.
- [x] Existing `HoldsFreshAbsoluteForMinimumFrames` regression test still passes.
- [x] `dotnet test` — 11481/11481 green.
- [x] Build deploys to KSP `GameData`. MD5 source = deployed; distinctive `ActiveReFlySessionInProgressProbe` UTF-16 string present in deployed binary.
- [ ] **In-game**: re-fly a multi-stage probe slot whose previous attempt staged before the re-fly's RP capture (predecessor "Kerbal X" or similar full-rocket recording activates immediately at re-fly load). The previous attempt's full-rocket ghost should appear at its recorded position at the activation UT, not 50–100 m further along its velocity vector. Compare against pre-fix behavior to confirm the slide is gone.


vl3c:
Closing — PR #824's last commit ('Fix FG recorder one-tick LLA staleness causing debris forward slide') fixes the same user-visible symptom upstream at the recorder. With #824 merged the activation-settle carve-out shipped here is no longer needed; in-game retesting on `main` @ c8e1646e confirms the ghost appears at the expected separation position with no further code changes.

PR #839, opened 2026-05-13
Skip OnFlightReady merge dialog while a Re-Fly session is active

vl3c:
## Summary

- Pressing Esc → Revert → **Retry from Rewind Point** during a Re-Fly was popping the tree merge/discard dialog on top of the fresh attempt the user just started. Retry became a two-click action (Retry, then "Discard Re-Fly Attempt" in the merge dialog) instead of the one-click "discard old, start fresh" the label promises.
- The OnFlightReady fallback that opens the dialog gated only on `HasPendingTree && !restoringActiveTree`. After `AtomicMarkerWrite` attaches the new fork to the pending tree and sets `ActiveReFlySessionMarker`, both gates pass — but the pending tree is owned by an in-progress Re-Fly, not a leaked auto-commit. The merge decision belongs to the scene-exit path (`SceneExitInterceptor`) once the attempt actually finishes.
- Same trigger applies to initial Re-Fly invocation, not just Retry. The Discard branch of `ReFlyRevertDialog` was already correct because it `LoadScene`s to `SPACECENTER`/`EDITOR` and never re-enters FLIGHT.

## Approach

- New pure helper `ParsekFlight.ShouldShowOnFlightReadyMergeDialog(hasPendingTree, restoringActiveTree, reFlySessionActive)` returns true only when a pending tree exists, no restore coroutine owns it (#293), AND no active Re-Fly session owns it.
- `reFlySessionActive` reuses the existing `ParsekScenario.IsReFlySessionActiveForQuickloadDiscard()` helper, which covers both the persisted marker and the `RewindInvokeContext.Pending` window before `AtomicMarkerWrite` recreates the marker.
- The OnFlightReady call site dispatches on the single decision; the Re-Fly skip path emits a dedicated Info log (`"… skipping merge dialog: active Re-Fly session owns the pending tree (Retry/initial invoke)"`). The #293 restoring-coroutine skip branch is preserved with its own Info log.

## Test plan

- [x] `dotnet test --filter OnFlightReadyMergeDialogGuardTests` — 7 / 7 pass (truth table: no-pending, canonical fallback, restoring-skip, Re-Fly-skip, both-skip, two no-pending short-circuits).
- [x] Full `dotnet test` — 11560 / 11561 pass; the 1 failure is the unrelated `InjectAllRecordings` test that requires KSP to be closed.
- [ ] Manual repro from the original log (`logs/2026-05-13_2049/KSP.log`): in flight in an active Re-Fly, press Esc → Revert → Retry from Rewind Point. After the RP load, the player should drop into a fresh attempt with **no popup**; the log should carry `[INFO][Flight] Pending tree '<name>' reached OnFlightReady — skipping merge dialog: active Re-Fly session owns the pending tree (Retry/initial invoke)` instead of the previous `[WARN][Flight] … showing tree merge dialog (fallback)`.
- [ ] Manual check that the merge dialog still fires when the Re-Fly attempt actually finishes (Recover / KSC / VAB) — via `SceneExitInterceptor`, not the OnFlightReady fallback.


vl3c:
Addressed both review suggestions in commit `21216c04`:

**Suggestion 1 (in-game test for the Retry path).** Extracted the OnFlightReady merge-dialog dispatch block into `ParsekFlight.MaybeShowPendingTreeMergeDialogOnFlightReady` so an in-game fixture can drive it directly via reflection without firing `GameEvents.onFlightReady`. New `OnFlightReadyMergeDialogGuardInGameTest`:
- `OnFlightReady_ActiveReFlySession_SkipsMergeDialog` arms a synthetic marker + pending tree, calls the extracted method, asserts no `ParsekMerge` popup spawns and `MergeDialogPending` stays false.
- `OnFlightReady_NoReFlySession_ShowsMergeDialog` is the positive control with the marker cleared; if the call site stops invoking the helper, this test still trips because the popup wouldn't appear.

**Suggestion 2 (`ShowDeferredMergeDialog` parallel hole).** Audited and left unchanged. Reasoning logged inline as a code comment at the dispatch site — short version:
- Every reachable call site (`OnSceneChangeRequested` at 1862 / 1888, `OnLoad` non-FLIGHT branch at 2211) fires in a non-FLIGHT scene, i.e. *after* the Re-Fly attempt has been concluded by the player's scene change. That is the correct moment for the merge decision, unlike `OnFlightReady` which fires the instant the user starts a fresh attempt.
- The 1-arg `MergeDialog.ShowTreeDialog` already detects `ActiveReFlySessionMarker != null` and renders the Re-Fly-specific message + suppressed-subtree closure ([MergeDialog.cs:111-180](Source/Parsek/MergeDialog.cs:111)). The dialog presented at site 5207 under an active marker is semantically correct.
- Skipping the dialog there would strand the user in a state with an unresolved Re-Fly and no way to commit/discard, which is strictly worse than showing the dialog.

Full suite green (11561/11561 including the previously-skipped `InjectAllRecordings`).

vl3c:
Confirmed and fixed the blocking issue in commit \`24da48f5\`.

**Trace from the original log** (`logs/2026-05-13_2049/KSP.log`):

1. **20:48:33.260** — Re-Fly probe destroyed during sess_0820. `ShowPostDestructionTreeMergeDialog` finalized the tree and stashed it as pending in **Finalized** state (line 321641-321644).
2. **20:48:40.767** — User pressed Esc → Revert → Retry from Rewind Point; \`OnSceneChangeRequested\` carried the Finalized pending tree across the LoadGame.
3. **20:48:41.382** — Post-load \`TryRestoreActiveTreeNode\` hit the #290d branch ("keeping in-memory Finalized tree … skipping .sfs replacement") and returned without setting \`ScheduleActiveTreeRestoreOnFlightReady\` (line 322542).
4. **20:48:43.38** — \`AtomicMarkerWrite\` attached fork \`rec_321b…\` to the Finalized pending tree and set the new marker.
5. **20:48:43.65** — \`OnFlightReady\`: \`restoreMode == None\` → \`TryRestoreCommittedTreeForSpawnedActiveVessel\` bailed on \`HasPendingTree\` → original code's merge-dialog fallback fired (line 323525); under the first-commit fix, the dialog was suppressed.
6. **No \`RestoreActiveTreeFromPending: resumed recording\` line ever fired for sess_322f** — compare 287999 for sess_0820's initial invocation, which did resume.

So the reviewer was exactly right: dialog gone, recorder also gone, "Retry" effectively did nothing.

**Fix (commit \`24da48f5\`):** Option (1) from the review — install the pending Re-Fly tree before the dispatch skips. Two new pure helpers:

- \`ShouldUpgradeRestoreModeForReFlyRetry(restoreMode, hasPendingTree, pendingTreeIsFinalized, reFlySessionActive)\` — upgrades \`restoreMode\` to \`Quickload\` in OnFlightReady when the Re-Fly retry pattern is detected (Finalized pending tree + active marker + no schedule already armed).
- \`ShouldAcceptFinalizedPendingTreeForReFlyRetry(hasPendingTree, pendingTreeIsFinalized, reFlySessionActive)\` — mirrors the same decision inside \`RestoreActiveTreeFromPending\`'s state gate so the coroutine accepts Finalized when called from this path.

The coroutine's existing marker-swap (\`ResolveInPlaceContinuationTarget\` + \`tree.ActiveRecordingId = markerSwap.TargetRecordingId\`) redirects the wait target to the new fork's vessel, and the post-match \`recorder.StartRecording(isPromotion: true)\` binds the live recorder.

The merge-dialog skip still does its job for the initial-invocation case — when the coroutine fires, \`restoringActiveTree=true\` is set synchronously at the coroutine's entry, so the dispatch hits the \`#293\` skip branch first (the Re-Fly-specific skip log only fires in the rare race where the dispatcher couldn't schedule the restore).

**Coverage:** 10 new xUnit cases in \`OnFlightReadyMergeDialogGuardTests\` cover both new helpers' truth tables. Full suite: 11571 / 11571 pass. The in-game test still asserts the merge dialog is suppressed; the recorder install is verified by the xUnit cases + the existing \`RestoreActiveTreeFromPending\` log assertions in \`RewindLoggingTests\`.

I'll need to do a manual KSP repro of the original log scenario before merge — the unit + in-game tests cover the decisions, but only a live Retry-from-RP exercises the full \`AtomicMarkerWrite → OnFlightReady → coroutine → recorder.StartRecording\` chain in one breath.

vl3c:
P2 addressed in commit \`139c162d\`.

Both safeguards from the review applied:

**(1) Skip when a real Re-Fly session is active.** Both \`OnFlightReady_ActiveReFlySession_SkipsMergeDialog\` and \`OnFlightReady_NoReFlySession_ShowsMergeDialog\` now precondition on \`scenario.ActiveReFlySessionMarker == null && !RewindInvokeContext.Pending\` and skip with an explicit "would corrupt live RenderSessionState" reason when either is set. Single-test runs don't get the batch flight baseline restore, so refusing to run is safer than trying to round-trip live session state across a synthetic-marker swap.

**(2) Re-align RenderSessionState in finally.** Both \`finally\` blocks now call \`Parsek.Rendering.RenderSessionState.RebuildFromMarker(savedMarker)\` right after restoring the marker. The skip check guarantees \`savedMarker == null\` here, so \`RebuildFromMarker(null)\` re-clears the anchor map (the canonical case post-control-test, where the body's \`Clear("ingame-test-control-setup")\` already wiped it; idempotent in the synthetic-marker test where nothing mutated render state). The explicit call documents the cleanup contract and stays correct if the precondition relaxes later.

Full suite: 11570 / 11571 pass — the lone failure is the unrelated \`InjectAllRecordings\` test that needs KSP closed (Directory.Delete refused because the save's KSP.log was locked). \`OnFlightReadyMergeDialogGuardTests\` (17) + \`Marker_NullAssignmentSites_AllPairWithClearCall\` (1) = 18 / 18.

vl3c:
Re-review (commit \`a891502b\`) — addressed three of the four suggestions:

**Suggestion 1 (placeholder-mode safety) — fixed.** Added \`ParsekScenario.IsReFlyInPlaceContinuationActive()\` (strict: marker set AND \`InPlaceContinuation == true\`) and narrowed both carve-outs to use it. Three behavioral cases now:

- **In-place continuation marker** (the user's bug case): both gates fire → restore coroutine schedules → recorder binds to the fresh fork → dialog skips. (Unchanged from before.)
- **Placeholder-mode marker** (PID changed or chain orphaned at \`AtomicMarkerWrite\` line 1096-1099): carve-outs do NOT fire (\`reFlyInPlaceContinuationActive=false\`), \`restoreMode\` stays None, original merge dialog fires as the recovery path. Pre-fix behavior preserved exactly where we know the recorder-restore cannot help.
- **Brief invoke window** (\`RewindInvokeContext.Pending=true\`, marker not yet written): the dialog-skip path uses \`InPlaceContinuationActive || Pending\` to avoid flickering the popup; the carve-outs do NOT use the \`Pending\` half because there is no marker to act on yet.

Params renamed \`reFlySessionActive → reFlyInPlaceContinuationActive\` across the three helpers, with an explicit truth-table case asserting the placeholder marker falls through to the dialog.

**Suggestion 2 (dispatcher/coroutine state re-read) — comment added.** Inline comment at the coroutine's \`acceptFinalizedForReFly\` computation explaining the intentional re-read independent of dispatcher snapshots. No code change beyond the comment — passing a dispatcher snapshot through would be overkill for a non-concrete race, and the explicit re-read keeps the coroutine self-contained.

**Suggestion 3 (in-game test restoringActiveTree check) — deferred.** Low-probability concern; would require reflection into the private field and complicate the test for marginal coverage. Skipping for now.

**Suggestion 4 (doc nit) — fixed.** The "fork attach not yet visible to OnLoad" phrasing was wrong (the race is at AtomicMarkerWrite-vs-OnFlightReady, not OnLoad — specifically the async flight-ready-deferred \`WaitForFlightReadyAndInvoke\` path where the marker write runs in a coroutine after \`OnFlightReady\` fires). Replaced with an accurate description and added a paragraph documenting the placeholder-mode handling.

Tests: 11572 / 11572 (added \`PendingTree_PlaceholderModeMarker_ReturnsTrue\` covering the new case).

Want to give it another Opus pass, or is this ready to merge?

PR #843, opened 2026-05-13
Trace-Sep: add interpolatedAnchorLocalDist field to PositionDebris rows

vl3c:
## Summary

- Adds a fourth distance field, `interpolatedAnchorLocalDist`, to the `Trace-Sep PositionDebris` log row that linearly interpolates the recorded anchor-local offset vector between bracketing-before and bracketing-after `frames[]` entries at `playbackUT` (vector lerp, then magnitude), instead of clamping to the bracketing-before entry's magnitude like the existing `recordedAnchorLocalDist` field does.
- The new value is computed by a pure static helper, `TraceSeparation.InterpolateAnchorLocalOffsetMagnitude`, with 12 unit tests covering sample hits, vector-then-magnitude midpoint contract, monotonic growth across an outward-moving bracket using the Kerbal X separation numbers from a recent investigation cycle, NaN for null/empty/single-frame/NaN-input cases, the right-edge clamp, and the duplicate-UT degeneracy fallback.

## Why

A recent investigation of a "the booster ghost is a bit too far from the parent at separation" report wrongly read the difference between `renderedParentDist` and `recordedAnchorLocalDist` as a rendering bug. The seed-only field stays constant across the entire 600 ms seed-to-first-sample bracket of a fresh debris recording, but the parent and debris really are separating during that window (parent thrusts forward, booster decelerates under drag, ~15 m apart at the seed and ~20 m apart at the next sample). The "drift" was mostly real physical separation that the recorder captured between the two `frames[]` endpoints. With `interpolatedAnchorLocalDist` in the row, a reader can compare:

- `renderedParentDist` ≈ `interpolatedAnchorLocalDist` ≈ `recordedBodyFixedDist` → playback reproduces recorded data faithfully (no rendering bug)
- `renderedParentDist` diverges from both recorded fields → playback bug
- The two recorded fields disagree → the two recording surfaces store inconsistent parent-vs-debris geometry

## No production behavior change

The new field is only written to the log row when the existing `traceEnabled && PlaybackWindowActive` gate is open, which requires both `ghostRenderTracing` to be on and a playback window opened by a first-debris-render trigger. Production runs pay nothing. The existing `recordedAnchorLocalDist` field is unchanged so existing log consumers keep working.

## Test plan

- [x] `dotnet test --filter "FullyQualifiedName~TraceSeparationTests"` — 20 / 20 pass (8 existing + 12 new).
- [x] `dotnet test` — full suite 11565 / 11565 pass, no collateral damage.
- [x] Manual diff review of the `PositionDebris` emit site to confirm no behavior change outside the log row.
- [ ] In-game smoke: enable `Settings → Diagnostics → Ghost render tracing`, fly a stage-separation in Watch mode with debris ghosts visible, grep `Trace-Sep PLAY [PositionDebris]` for the new `interpolatedAnchorLocalDist=` token and confirm it evolves across the first bracket while `recordedAnchorLocalDist` stays flat.

## Doc updates in the same commit

- `CHANGELOG.md` — Internals + Tests entries under 0.9.2.
- `docs/dev/todo-and-known-bugs.md` — observability bullet for the playback-side `PositionDebris` row updated to mention the new field, and the seven-question diagnostic walk's step 7 now compares `renderedParentDist` to `interpolatedAnchorLocalDist` rather than `recordedAnchorLocalDist` (which conflates real physical separation with rendering error inside a wide bracket).


vl3c:
Addressed the two should-fix items from the Opus review pass in e3ee5f44:

1. **Bracket-finder semantics.** Rewrote the helper's loop to use the same early-break pattern as `ParsekFlight.cs`'s existing `recordedAnchorLocalDist` loop (break on first `frames[j].ut > playbackUT`), and dropped the separate `afterIdx` scan in favor of `beforeIdx + 1`. The two trace fields now pick the same bracket by construction on any UT-monotonic input. Documented the monotonic-input assumption in the XML doc so future callers know the contract.

2. **Duplicate-UT test pinned.** The test now asserts the exact deterministic outcome (18.788, the last sample's magnitude via the right-edge clamp path). The previous "not-NaN and > 0" was too weak to catch a regression that flipped the duplicate handling.

Plus two nice-to-have test gaps Opus flagged:

- `PositiveInfinity playbackUT → NaN` — parity with `ComputeLerpAlpha`'s Infinity guard.
- `NaN-UT in frame → no crash` — pins the contract that a malformed frame UT falls back to the bracketing-before magnitude via the existing NaN-alpha guard.

All 22 TraceSeparation tests pass; full suite 11567/11567 pass. The interpolation contract and the `PositionDebris` log emit are unchanged from the first commit — this is a contract-tightening + test-strengthening pass.

vl3c:
Fixed both stale-doc items in d0f6548c (comment- and CHANGELOG-only, no code change):

- **ParsekFlight.cs:17368-17386** — the `recordedAnchorLocalDist` comment no longer instructs readers to compare it against `renderedParentDist` as a fidelity check (that was the seed-only interpretation this PR is replacing). The comment now directs readers to `interpolatedAnchorLocalDist` for fidelity and keeps `recordedAnchorLocalDist` as the bracket-left-edge snapshot view, still useful for direct comparison against the section's `bodyFixedFrames[]` bracketing-before sample via `recordedBodyFixedDist`.
- **CHANGELOG.md** — dropped the "falls back to the bracketing-before magnitude on degenerate zero-span brackets" line (that was the helper's pre-rework duplicate-UT behavior; the rework moved duplicate-UT into the right-edge clamp path). The entry now describes the actual current behavior and adds the bracket-finder-matches-existing-loop note.

No code change; build still passes.

vl3c:
Fixed in 2746aa16 — the Tests-section CHANGELOG line now describes the actual current contract (last-sample clamp on the duplicate-UT bracket-collapse case) instead of the pre-rework bracketing-before fallback, and adds the two gap tests from the followup (PositiveInfinity playback-UT NaN, NaN-UT-in-frame skip-vs-dereference). Doc-only change.

PR #844, opened 2026-05-13
Skip deferred merge dialog while a Re-Fly session is active

vl3c:
## Summary

PR [#839](https://github.com/vl3c/Parsek/pull/839) added a Re-Fly guard on the `OnFlightReady` merge-dialog fallback. The post-PR Opus review flagged the same shape of leak on the symmetric **exit-side** fallback `ParsekScenario.ShowDeferredMergeDialog` (the coroutine the `OnLoad` fallback queues when `SceneExitInterceptor`'s pre-transition `HighLogic.LoadScene` prefix misses a FLIGHT->non-FLIGHT transition: mod compat, KSP version drift, foreign LoadScene patch).

All three call sites in `ParsekScenario.OnLoad` (`~1862` / `~1888` / `~2211`) are gated by `HighLogic.LoadedScene != GameScenes.FLIGHT`, which is exactly the missed FLIGHT->KSC/TS/MAINMENU case. In that miss case the Re-Fly marker survives the scene change (because the SceneExitInterceptor `ReFlyAttempt` dialog never fired), and the deferred coroutine would have opened `MergeDialog.ShowTreeDialog` over a pending tree owned by the active Re-Fly attempt.

- New pure helper `ParsekScenario.ShouldShowDeferredMergeDialog(hasPendingTree, reFlySessionActive)` mirrors `ParsekFlight.ShouldShowOnFlightReadyMergeDialog`.
- The coroutine now consults `IsReFlySessionActiveForQuickloadDiscard()` after the 60-frame wait (right after the existing "pending consumed during wait" guard) and bails with an Info skip line whose wording matches the OnFlightReady skip line. `mergeDialogPending` is cleared on the skip so subsequent dispatches are not blocked.
- xUnit `DeferredMergeDialogGuardTests` covers the four-quadrant truth table for the new helper.

## Test plan

- [x] `dotnet build Source/Parsek/Parsek.csproj` (0 warnings, 0 errors)
- [x] `dotnet test Source/Parsek.Tests/Parsek.Tests.csproj` (11590 passed, 0 failed)
- [x] `dotnet test --filter "FullyQualifiedName~DeferredMergeDialogGuardTests"` (4 passed)
- [ ] In-game smoke: not runnable without a live KSP repro of the SceneExitInterceptor-missed path; the pure helper truth table plus the existing OnFlightReady in-game test cover the parallel behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to Parsek are documented here.
 
 ### Bug Fixes
 
+- The deferred merge-dialog fallback that fires when `SceneExitInterceptor`'s pre-transition prefix misses a FLIGHT->non-FLIGHT transition (mod compat, KSP version drift, foreign LoadScene patch) now skips when an active Re-Fly session owns the pending tree. Mirrors the OnFlightReady fallback guard so the same shape of leak does not reach the deferred coroutine path.
 - Planted flags now survive Re-Fly. The post-load strip used to delete flag vessels along with other unmatched siblings, silently erasing the FlagPlant career milestone; the stripper now preserves any `VesselType.Flag` vessel regardless of slot-map membership.
 - Re-Fly now re-spawns prior-committed sibling vessels (e.g. a landed capsule from an earlier mission) at their terminal endpoint instead of leaving them permanently un-materialized after the Re-Fly load strip removes them from the save.
 - Re-Fly fork ghosts no longer flame their booster engines at full throttle when the recorded engine was shut down. Chain-promotion seed emission now scopes narrowly to engine sentinels, fires only when the active recording lacks any engine events, and anchors the sentinel UT at the recording's StartUT when the recording has real trajectory data (including sandbox-epoch starts at UT 0) and current UT for genuinely fresh forks, so the ghost playback orphan-engine heuristic stays correct without poisoning boring-tail trim.
--- a/Source/Parsek.Tests/DeferredMergeDialogGuardTests.cs
+++ b/Source/Parsek.Tests/DeferredMergeDialogGuardTests.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Parsek.Tests
+{
+    /// <summary>
+    /// Unit tests for <see cref="ParsekScenario.ShouldShowDeferredMergeDialog"/>.
+    /// The deferred-merge-dialog fallback opens the tree merge dialog only when
+    /// there is a pending tree AND no active Re-Fly session owns it. Mirrors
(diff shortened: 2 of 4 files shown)


vl3c:
Closing after an independent Opus review caught two issues that make this PR a no-op at best and a regression at worst.

**P1.1 — PR #839's review-followup commit already audited this site and rejected the guard.**

Commit `21216c04` ("Add in-game test for OnFlightReady Re-Fly guard wiring") on the `fix-refly-retry-merge-dialog` branch added an audit comment directly above `MergeDialog.ShowTreeDialog(RecordingStore.PendingTree)` in `ShowDeferredMergeDialog`:

> "unlike the OnFlightReady fallback ... this deferred coroutine does NOT need the active-Re-Fly skip guard. The OnFlightReady fallback fires the moment the player enters FLIGHT after Re-Fly invocation or Retry-from-RP — i.e. when the user just started flying a fresh attempt and the dialog would be wrong-timing. The deferred coroutine, by contrast, only fires in a non-FLIGHT scene (call sites at 1862, 1888, 2211 all gate on leaving / having left FLIGHT), which means the Re-Fly attempt is already concluded by the player's scene change. Surfacing the merge decision there is the correct recovery path when SceneExitInterceptor missed the pre-transition catch. MergeDialog.ShowTreeDialog already detects ActiveReFlySessionMarker != null and renders the Re-Fly-specific message + suppressed-subtree closure, so the dialog presented here is semantically correct."

Verified `MergeDialog.cs:111-180`: the 1-arg `ShowTreeDialog` consults `ActiveReFlySessionMarker` and renders the Re-Fly-flavored short message + session-suppressed-subtree closure. The non-FLIGHT scene gate means the player has already concluded the attempt. The audit's reasoning is correct.

That follow-up commit is on `fix-refly-retry-merge-dialog` (the branch behind PR #839) and was not yet on `main` when this PR was authored, which is why the original task didn't see it.

**P1.2 — The guard introduces a worse failure mode.**

If the skip fires, `mergeDialogPending = false` but the pending tree + Re-Fly marker remain attached. On the next `OnSave` with `LoadedScene != FLIGHT`, `SafetyNetAutoCommitPending` (`ParsekScenario.cs:606-625`) matches `!mergeDialogPending` and calls `AutoCommitTreeGhostOnly` + `CommitPendingTreeAsApplied`. That bypasses the supersede / tombstone / `MergeJournalOrchestrator` machinery that owns Re-Fly merge correctness, silently ghost-committing the Re-Fly fork. Strictly worse than the pre-PR behavior of opening the Re-Fly-flavored dialog and letting the player decide.

**Verdict:** the in-flight `MergeDialog.ShowTreeDialog` Re-Fly-aware rendering is the correct UI here; the FLIGHT->non-FLIGHT exit case is not an "OnFlightReady-style leak." Closing without merging.

PR #17, opened 2026-03-07
Fix infinite science duplication via recording merge/revert

vl3c:
## Summary

- **Bug**: After reverting, `LastAppliedResourceIndex` resets to -1, causing science deltas to be fully re-replayed via `AddScience()`. Since `AddScience()` doesn't update `ScienceSubject` records, experiments appear undone after revert, allowing infinite re-earning of the same science.
- **Fix**: Track per-experiment science subjects via `GameEvents.OnScienceRecieved` during recording. On commit, merge into a persistent committed store (max-wins for partial experiments). Harmony postfix patches on `ResearchAndDevelopment.GetExperimentSubject` and `GetSubjectByID` inject committed science values, making KSP's diminishing returns formula compute correct remaining science.
- Added `ClearScienceSubjects()` to the Wipe All path so committed subjects don't persist after a full data wipe
- 8 new unit tests covering commit logic, max-wins semantics, clear, and edge cases

## Changed files

| File | Change |
|------|--------|
| `Patches/ScienceSubjectPatch.cs` | New Harmony postfix patches on R&D subject lookup |
| `GameStateEvent.cs` | Added `PendingScienceSubject` struct |
| `GameStateStore.cs` | Committed science dict, commit/query/clear/serialize/deserialize |
| `GameStateRecorder.cs` | `OnScienceRecieved` subscription, pending subject capture |
| `RecordingStore.cs` | Wired commit/discard of pending subjects |
| `ParsekUI.cs` | Added `ClearScienceSubjects()` to Wipe All |
| `GameStateEventTests.cs` | 8 new tests for committed science subjects |

## Test plan

- [x] `dotnet build` — 0 errors
- [x] `dotnet test` — 1204 passed, 1 skipped, 0 failed
- [x] In-game: record mission with science transmissions, revert, verify experiments show reduced/zero remaining science
- [x] In-game: partial science (transmit vs recover) — verify remaining science is correctly computed
- [x] In-game: Wipe All — verify experiments become re-earnable after wipe

PR #871, opened 2026-05-16
Plan: segment-scoped auto-record for stock Fly / Switch-To buttons

vl3c:
## Summary

- Design doc for a v0.9.2 follow-up to PR #866: stock Fly / Switch-To buttons should immediately auto-start a segment-scoped continuation recording with a distinct ID, so Merge commits only the segment and Discard preserves committed history.
- Three confirmed stock UI sources (verified via decompilation of `Assembly-CSharp.dll`): Tracking Station Fly, KSC nearby-vessel marker Fly, and in-flight Map view Switch To. `[`/`]` keyboard cycling and other generic focus changes stay on the existing first-modification watcher.
- Plan only — no code changes. Adds [docs/dev/plans/segment-scoped-switch-fly-autorecord.md](docs/dev/plans/segment-scoped-switch-fly-autorecord.md) and a corresponding entry in [docs/dev/todo-and-known-bugs.md](docs/dev/todo-and-known-bugs.md).

## Plan highlights

- New `Recording.SwitchSegmentSessionId` field — explicitly NOT a reuse of `CreatingSessionId` (which is load-bearing in `LoadTimeSweep` / `MergeJournalOrchestrator` / `MergeDialog.TryDiscardActiveReFlyAttempt` / provisional-recording detection).
- New non-claiming `BranchPointType.VesselSwitchContinuation` enum value with a format-version bump and consumer audit list.
- Three narrow Harmony targets: `SpaceTracking.FlyVessel`, `KSCVesselMarkers.FlyVessel`, and `MapContextMenuOptions.FocusObject.OnSelect` (postfix-arms-on-success, not prefix). Explicitly forbids patching `FlightGlobals.SetActiveVessel` / `ForceSetActiveVessel` / `FlightDriver.StartAndFocusVessel`.
- Marker-lifetime asymmetry spelled out: TS/KSC cross-scene → full `OnSave`/`OnLoad` round-trip; Map Switch-To in-scene → ephemeral, ~1-frame consume from `OnVesselSwitchComplete`. Uniform marker struct, differentiated consume hooks.
- Four named #866 suppression sites enumerated and narrowed for marker-owned new IDs; clone-parent sidecar interaction spelled out.
- Two-rapid-switches semantics, intent TTL (10s TS/KSC, 2s Map), ERS/ELS routing note, Board/Dock composition during active segment, dialog copy templates by entry reason.
- Resolved open decisions (4 + 1 escape hatch) recorded in the plan.

## Test plan

- [x] Plan reads cleanly end-to-end and the corrected decompilation findings agree with the implementation order.
- [x] Implementation order step 1 (this PR) lands the design before any code follows.
- [ ] Subsequent implementation PRs cite this plan and start at step 2 (`SwitchSegmentSession` marker + `Recording.SwitchSegmentSessionId` field with serialization tests).


vl3c:
Addressed the review in `eac6cb2e`. Item-by-item:

### Must-fix items

1. **Prefix vs Postfix contradiction + arming-timing risk** — Decompiled `FlightGlobals.setActiveVessel` to settle it. `GameEvents.onVesselChange.Fire(v)` runs *synchronously* inside the method body before `return true`, so Parsek's `OnVesselSwitchComplete` listener fires before `SetActiveVessel` returns, which is before any Postfix on the calling `FocusObject.OnSelect` would run. Postfix-arms-on-success is therefore wrong — it would arm the marker too late. The plan now specifies **Prefix arms + Postfix clears on refusal**, scoped by an `IntentId` passed via `__state`. The decompiled early-return paths (vessel null, vessel == active, `ClearToSave` fail × 6, unowned discovery) and the unloaded-vessel branch are enumerated as the cases the Postfix cleanup catches. Implementation Order step 7 rewritten to match.

2. **Format-version bump vs no-migration policy** — Reconciled. Per `memory/feedback_no_recording_compat.md`, pre-1.0 we reject old saves at load rather than migrating. The plan now says explicitly: the format bump is real, but the test obligation is forward-compat / `default:`-arm hygiene in `GhostingTriggerClassifier` consumers (and any other `switch (bp.Type)` site). Dropped the "load old tree with unrecognized integer" test; kept a new-format round-trip test only.

3. **`autoRecordOnExplicitVesselSwitch` default** — Now states explicitly **default ON**. The Resolved Decision #5 escape hatch only flips it OFF if playtest finds Map Switch-To ping-pong noisy.

4. **`EntryReason.RecoveredConfirmedIntent`** — Dropped from the enum. Replaced the missed-switch-recovery paragraph with a clear "treat as cross-run-orphaned and clear" rule. The `OnLoad` tail clears stale serialized markers with `stale-cross-run` when the loaded scene is the source scene (player loaded a save where they had armed a marker but never transitioned to FLIGHT in that session).

### Smaller concerns

- **`expected-bg-entry-missing` log** — Added to the parent-boundary flush helper section. When the focused vessel is not in the BG map at segment-creation time (TS Fly / KSC Fly cases where the vessel just came live), log the case explicitly instead of silently skipping the flush.
- **Cancel between dialogs** — Added: second dialog offers Cancel alongside Merge / Discard; Cancel returns the player to FLIGHT with the segment-pruned pending tree intact. Scene exit is blocked only until the player picks one of three. Mirrors the existing `MergeDialog.MergeDiscard` input-lock contract.
- **Cross-run orphan marker** — Already noted above (item 4): explicit `stale-cross-run` clear in the `OnLoad` tail. TTL would eventually catch it but the deterministic load-time clear avoids surprises.
- **CHANGELOG** — Plan-only PR, no behavior ships; subsequent implementation PRs will add CHANGELOG entries per the per-commit doc rule.
- **Test 22 clarification** — Added: this is an in-game test in `InGameTests/RuntimeTests.cs` (FLIGHT scene category), not a unit test. `FocusObject.OnSelect` is inaccessible from the xUnit assembly.

All "good" items from the review are retained as-is. Plan now 459 lines; commit pushed.

vl3c:
Re-review addressed in `50647b80`. Item-by-item:

### From the first review (concerns A-D)

- **A. `stale-cross-run` detection underspecified** → Added `ProcessSessionId` to the marker fields: a non-serialized GUID created once at scenario init, captured into the marker at arm time, compared against the current process's GUID on `OnLoad`. Mismatch ⇒ deterministic `stale-cross-run` clear; match ⇒ same-process F5/F9, leave armed. The three resume cases (same-process F5, same-process load-into-armed, fresh-process) are spelled out explicitly.
- **B. Map Switch-To → unloaded-vessel branch outcome** → Rewrote the unloaded-vessel paragraph: the deliberate outcome is that Map Switch-To to an unloaded vessel does NOT immediate-start; the player falls through to the first-modification watcher. No "rearm fresh" hand-wave; the path is named as an acceptable degradation. Added in-game test #12 to cover it.
- **C. Prefix-on-Prefix race** → Added explicit `stale-intent-superseded` log to the FocusObject Prefix when it sees a still-armed marker with a different IntentId. Tied back to the existing "Two rapid switches" reason name.
- **D. Setting gate missing from Prefix predicate** → The Prefix predicate in step 7 now lists `ParsekSettings.AutoRecordOnMapSwitchTo` as the first AND-gate. Steps 5 (TS Fly) and 6 (KSC Fly) similarly gate on `AutoRecordOnTsFly` and `AutoRecordOnKscFly`.

### From the second review (concerns 1-7)

- **1. Goal 4 entry-path missing** → Added a dedicated "Map Switch-To to a vessel that is a tracked background member of the current uncommitted active tree" subsection in §"Behavior by Entry Path". Spells out the 7-step path (consume → background prior → look up BG recording → flush BG parent → walk to terminal leaf → attach `VesselSwitchContinuation` → arm session). Notes this case only applies to Map Switch-To (TS/KSC have already stashed/committed by scene-cross).
- **2. Setting granularity** → Replaced the single `autoRecordOnExplicitVesselSwitch` toggle with three per-source toggles (`autoRecordOnTsFly`, `autoRecordOnKscFly`, `autoRecordOnMapSwitchTo`), all defaulting ON. Resolved Decision #2 and Decision #5 updated to match — the Map escape hatch now disables only Map Switch-To, not the other two.
- **3. Test #17 split** → Split into 17a (same-process F5/F9 within session — marker survives, consumed normally) and 17b (cross-run load on fresh process — marker clears with `stale-cross-run` on `OnLoad`).
- **4. Decompiled snippet as `///` comment** → Step 7 now mandates including the actual decompiled body of `FlightGlobals.setActiveVessel` (or the load-bearing 30-line excerpt covering `onVesselSwitching.Fire` → `MakeActive()` → `onVesselChange.Fire`) as a `///` comment block above the patch class, so the synchronous-firing claim is self-documenting.
- **5. Steps 5/6 under-specified for refusal handling** → Both steps now explicitly state that no Postfix-refusal handling is required because `SpaceTracking.FlyVessel` and `KSCVesselMarkers.FlyVessel` don't return a synchronous bool; consume happens on the next FLIGHT `OnLoad`. Aborted FLIGHT loads are covered by the `ProcessSessionId` cross-run-orphan mechanism + wall-clock TTL.
- **6. Dialog copy substitution mechanism** → Specified C# interpolation `$"..."` (matches existing `MergeDialog.cs` style) with `{vesselName}` as shorthand for `{rec.VesselName ?? "this vessel"}`.
- **7. "Focus" naming confusion** → Reworded the Summary's Map-view note and the patch-shape bullet: the constructor passes the literal string `"Focus"` to the base `MapContextMenuOption(string)` ctor (a string identifier, NOT a C# class name). The C# type is `MapContextMenuOptions.FocusObject`. The "Decompilation correction" paragraph got the same clarification.

### Also resolved while in there

- **`ShouldDeferPendingEventMilestoneFlushForSave` narrowing choice** (open from prior round) — Picked option 1: allow marker-owned new-ID milestone flushes once the marker is durable. The predicate now distinguishes same-ID committed-restore tails (deferred) from marker-owned new-ID events (flushable). Option 2 (separate milestone-shadow file) rejected as more code/invariants for a rare case. Test obligation pinned: marker-owned new-ID events appear in milestone file after F5/save/reload and are still purged on Discard.

Plan now 478 lines; pushed.

vl3c:
Follow-up review addressed in `3c1a0b34`. Three items, all fixed:

1. **Goal 4 entry path — "TS Fly and KSC Fly cannot land here" overstates.** Removed the overstatement. The section is now titled "Fly / Switch-To a vessel that is a tracked background member" (not "Map Switch-To to..."), with a paragraph explicitly stating that routing depends on what the focused vessel matches at consume time, not on which UI button armed the marker. The cross-session edge case is named: F5 mid-flight with uncommitted active tree → resume save later → subsequent TS/KSC Fly into a BG member of that restored tree routes through this branch (matched recording is in a pending tree, not committed storage). Picks one branch and owns it — Goal 4 branch wins for any BG match in active/pending tree, regardless of source.

2. **Goal 4 step 3 lookup-failure fallthrough.** Step 3 now spells out the failure modes (orphaned recording in tree with no BG entry, BG entry without an in-tree recording shell, PID collision, restore-time tree-state corruption) and the fallthrough: drop into the unrelated-vessel branch, start a standalone segment, log `bg-lookup-failed-start-standalone` with the focused PID and reason. Explicit "do not raise; the immediate-start path must degrade gracefully."

3. **Decompiled-body `///` comment hygiene extended to steps 5 and 6.** Both now include the requirement to embed the load-bearing decompiled excerpt above their patch class — `SpaceTracking.FlyVessel` for step 5 (documents void return + `FlightDriver.StartAndFocusVessel` call shape), `KSCVesselMarkers.FlyVessel` for step 6 (same). Step 7 (the original Map patch) gets a closing sentence promoting the rule to all three sites: "every patch whose correctness depends on an assumption about a stock method's body should carry the excerpt that proves the assumption, so the assumption ages with the patch."

Plan now 488 lines; pushed.

vl3c:
Addressed in `02326a8a`. Three items:

1. **`ProcessSessionId` static lifetime hazard.** The "ParsekScenario.Awake" hint was wrong — ScenarioModule reinstantiates on every scene transition, so an Awake-based initializer would regenerate the GUID on every scene change and defeat the mechanism entirely. The plan now mandates a `static readonly Guid ProcessSessionId = Guid.NewGuid();` field initializer on a long-lived static class (suggested home: a small new `ParsekProcess` helper, or reuse the already-long-lived `ParsekLog`). Explicit "do NOT initialize from `MonoBehaviour.Awake`" warning. The static field initializer runs once per AppDomain on first type access — the right lifetime.

2. **No test for Goal-4 entry path.** Added Test #23 and #24:
   - #23: Map Switch-To to a BG-tracked vessel of the live uncommitted active tree — close BG via `OnVesselRemovedFromBackground`, attach `VesselSwitchContinuation` under BG terminal leaf, BG recording's `RecordingId` unchanged (no in-place promotion). Covers both no-existing-continuation and existing-continuation cases.
   - #24: BG-lookup failure fallthrough — each of the four named failure modes (orphan recording, BG without shell, PID collision, corrupted restored tree) gets its own case asserting `bg-lookup-failed-start-standalone` log + standalone segment.

3. **Same-process old-save wording overstated.** Rewrote that bullet. The safety isn't "the player was about to Fly" — it's the TTL clearing the marker on `OnLoad` if more than 10s elapsed (common case), or `stale-intent-superseded` overwriting it within the 10s window when the player clicks Fly. Neither case produces an incorrect immediate-start.

Plan now 490 lines.

PR #873, opened 2026-05-16
Fix splitter empty-HEAD on recording whose data starts at rewindUT (Pass 7)

vl3c:
## Summary

Fixes a regression discovered while playtesting PR #872: a Re-Fly on a recording whose sampled content starts **exactly at the rewind UT** (but whose `ExplicitStartUT` carries an earlier stale value) silently produces an empty HEAD that survives the supersede write-set and shows up as a **phantom STASH entry** on the slot. This PR is targeted at `fix-supersede-identity-scope` so it lands before #872 merges.

## Repro

Live save with two stages: capsule (`77d213d7…`, crashed) + probe (`0aeaa94d…`, also crashed, Points start at the section boundary `82.96`, `ExplicitStartUT=82.4` from an earlier optimizer pass).

1. User clicks Re-Fly on the probe slot (rewindUT = 82.96).
2. Re-flown probe crashes again; user commits via the merge dialog.
3. `MergeJournalOrchestrator` invokes `RecordingTreeSplitter.SplitOriginAtRewindUT`.
4. Strict-span check passes because `origin.StartUT (82.4) < rewindUT (82.96) - epsilon`. But the **actual data** starts at `82.96`.
5. `RecordingOptimizer.SplitAtSection` partitions on `TrackSections[0].startUT = 82.96` (not `ExplicitStartUT`). HEAD ends up with `[0.00..0.00]`, 0 Points / 0 sections; TIP gets the full content `[82.96..92.72]`.
6. `IsPreRewindCarveOut` correctly identifies that empty HEAD as a `PreRewindChainHead` (it satisfies the predicate trivially — empty bounds + chain-shape match) and excludes it from the supersede write-set.
7. HEAD remains visible in ERS as an unfinished-flight zombie (`terminalState=Destroyed` + slot 1) — third phantom STASH entry where only two were expected.

The user observed STASH carrying capsule + phantom empty probe HEAD + re-flown probe = 3, instead of capsule + re-flown probe = 2.

## Fix

The strict-span check at [`RecordingTreeSplitter.cs:482-508`](https://github.com/vl3c/Parsek/blob/fix-splitter-empty-head/Source/Parsek/RecordingTreeSplitter.cs#L482) now reads from `Recording.TryGetActualTrajectoryBounds`, which derives bounds strictly from sampled content (Points / OrbitSegments / playable TrackSections) and **excludes `ExplicitStartUT` / `ExplicitEndUT`**. A recording whose actual data starts at or after `rewindUT` falls through to whole-recording supersede instead of producing the bogus empty HEAD.

The skip log reports both views (actual bounds + blended bounds) so an operator triaging "phantom STASH entry after Re-Fly" can spot stale `ExplicitStartUT` at a glance.

`TryGetActualTrajectoryBounds` is promoted from `private` to `internal` so the splitter and tests can call it.

## Why this belongs in #872

The regression is created by #872's own splitter (the empty HEAD) and protected by #872's own carve-out (the chain-head case). Letting #872 merge with the regression present would ship a brand-new user-visible bug — phantom STASH entries every time someone Re-Flies a recording whose `ExplicitStartUT` is stale relative to its first sample. The change is narrow (~40 lines of logic + 1 visibility bump + 1 new test) and lives entirely in code introduced by #872.

## Test plan

- [x] Full xUnit suite passes (11872 tests, +1 new regression test).
- [x] `SplitOriginAtRewindUT_StaleExplicitStartUT_NoEmptyHeadSplit` pins the exact repro: Points at `[82.96, 92.72]`, `ExplicitStartUT=82.4`, `rewindUT=82.96` → splitter returns `Skipped=true` / `SkipReason=OriginDoesNotSpanRewindUT`, no TIP created, origin payload untouched, skip log mentions both actual and blended bounds.
- [x] Existing `OriginEntirelyPreRewind` / `OriginEntirelyPostRewind` skip cases still pass — the stricter check doesn't regress the symmetric skip paths.
- [ ] Manual: re-run the user's repro against the build with this PR and verify the STASH count drops from 3 to 2.


vl3c:
## Follow-up commit: 6ad9e1fd

Applied the Opus review's Medium findings, both Low test-coverage items, all the nits, and the external review's cosmetic skip-log reframing. Full xUnit suite green at 11877 (was 11872).

**Code changes in 6ad9e1fd:**

- `IsPreRewindCarveOut` debris + chain-head cases now read bounds from `Recording.TryGetActualTrajectoryBounds` instead of the blended `Recording.StartUT/EndUT` view. This is the consumer-side defense in depth — closes the bug class even for legacy saves or hypothetical future producer regressions that ship an empty HEAD with chain shape matching TIP.
- Splitter's Step-2.7 debris reparent walk gets the same treatment. A debris recording with actual content post-rewind but `ExplicitStartUT = parent.branchUT` would otherwise stay parented to HEAD when it should follow TIP.
- Splitter skip-log message reframed to cover both "launch row will be hidden" and the no-content subset where whole-recording supersede is the correct outcome.
- Splitter comment now references the actual findable grep target in `todo-and-known-bugs.md` ("Pass 7 review fix").
- `actualBoundsStr` switched from concat to `$"..."` interpolation.
- CHANGELOG entry's trailing sentence trimmed.

**Test additions:** 5 new (`SplitOriginAtRewindUT_StaleExplicitEndUT_NoEmptyTipSplit`, `SplitOriginAtRewindUT_NoActualBounds_NoSplit`, `IsPreRewindCarveOut_NoActualTrajectoryBounds_NotCarvedOut`, `IsPreRewindCarveOut_StaleExplicitStartUT_DebrisDoesNotEscape`, `IsPreRewindCarveOut_StaleExplicitEndUT_GenuineHeadStillCarvedOut`). 10 existing carve-out fixtures gained a `StampActualBounds` helper so their Explicit-only recordings now also carry minimal Points matching the Explicit bounds — test intent unchanged, but they now exercise the new TryGet-based cutoff comparisons.

---

**External review item 1 — "Why can ExplicitStartUT end up stale?"** Investigated; **not a regression** and no upstream fix needed.

`RecordingOptimizer.SplitAtSection` does call `StampExplicitBoundsFromPayload` after a split, so the divergence isn't from a missing restamp there. Looking at the other writers (e.g. [ParsekFlight.cs:5910](https://github.com/vl3c/Parsek/blob/fix-splitter-empty-head/Source/Parsek/ParsekFlight.cs#L5910), [BackgroundRecorder.cs:748](https://github.com/vl3c/Parsek/blob/fix-splitter-empty-head/Source/Parsek/BackgroundRecorder.cs#L748)) the pattern is `ExplicitStartUT = branchUT` — set at child-recording creation to record the **logical / intended** start UT. The sampler typically catches up a frame or two later, so `Points[0].ut > ExplicitStartUT` is the normal case for any branched recording. This is by design (and load-bearing for downstream consumers that want to see a contiguous timeline even when the sampler skipped a frame on branch).

So the bug class isn't "stale ExplicitStartUT" in the maintenance-debt sense — it's "the splitter / carve-out predicates were using the wrong contract (logical bounds) for the decisions they needed to make (content bounds)". Pass 7 + this commit address that at every consumer site I could find. No separate audit ticket needed.

**External review item 2 — Cosmetic skip-log reframing.** Done in 6ad9e1fd — covered above.

PR #874, opened 2026-05-16
Prefer non-debris as CoBubble primary so re-fly controlled-child ghosts don't snap onto sibling debris

vl3c:
## Summary

- Short-term fix for the visual seam reported on the PR #872 build: during Re-Fly the Kerbal X Probe ghost (the booster lower stage, a controlled vessel) visibly snapped mid-flight onto a sibling radial-booster debris piece's trajectory at UT 34.74 and crossfaded back at UT 38.76 when the debris crashed.
- Adds Rule 6 to `CoBubblePrimarySelector.SelectPrimaryForPair` (fires between rules 2 and 3 in evaluation order): when one side of a CoBubble pair has `Recording.IsDebris == true` and the other does not, the non-debris side becomes primary. Controlled vessels (probes, landers, capsules) now play their own standalone Absolute trajectory end-to-end; only debris peers ride through a controlled primary via the saved offset.
- The deeper recorder-side fix — giving controlled-decoupled children a `DebrisParentRecordingId` parent-anchored surface like debris already has, eliminating dependence on CoBubble peer-blend for this case — is filed as a new Open entry in `docs/dev/todo-and-known-bugs.md` for the next release cycle. PR #873 stays as the safety-net guard for cross-tree co-bubble formations the recorder fix won't cover.

## Repro

From `logs/2026-05-16_2010_pr872-kerbalx-ghost-switch/KSP.log`:

| UT | Wallclock | Event |
|---|---|---|
| 24.54 | 20:06:59.583 | Probe ghost #7 appears, `activeFrame=Absolute sectionUT=24.26-41.72` (correct) |
| 34.74 | 20:07:09.990 | `Pipeline-CoBubble Blend window enter: peer=b11ef3d4(probe) primary=fa429137(debris)` — probe snaps onto debris track |
| 38.76 | 20:07:14.040 | `Blend window exit: ... reason=crossfade-tail` — debris crashed at UT 38.8, trace EndUT clipped |
| 41.72 | 20:07:16.969 | Probe terminal, explosion FX |

Selector log shows `rule=3` (earlier `StartUT`): debris 20.66 < probe 24.26, so the older sibling-debris recording was picked as primary against the newer controlled-probe peer.

## Rule order after this change

1. Live wins
2. Closest-to-live in DAG ancestry
3. **NEW: non-debris over debris** (rule index 6 to preserve §10.1 numbering of the original five)
4. Earlier StartUT (was rule 3)
5. Higher sample rate (was rule 4)
6. Ordinal id tiebreaker (was rule 5)

## Test plan

- [x] `dotnet build` clean
- [x] `dotnet test` — all 11,803 tests pass
- [x] Three new xUnit tests in `CoBubbleBlenderTests`:
  - `PrimarySelection_NonDebrisBeatsDebris_OverridingEarlierStartUT` — canonical fix assertion
  - `PrimarySelection_LiveStillBeatsNonDebrisRule` — Rule 1 still beats Rule 6 when the debris side is live
  - `PrimarySelection_BothDebris_FallsThroughToLaterRules` — pairs of two debris fall through to Rule 3 unchanged
- [ ] In-game verification by replaying the PR #872 Kerbal X repro: Re-Fly the upper stage and confirm the lower-stage probe ghost plays smoothly end-to-end with no mid-flight snap, regardless of whether sibling debris pieces crash during its playback window.

## Notes

- Selector-only change. `CoBubbleOverlapDetector` is unchanged, so existing `.pann` co-bubble traces stay valid and no schema bump is needed.
- Rule index 6 is non-sequential by design — keeps the original §10.1 rule indices 1-5 stable so historical log lines and debug references don't need updating. The docstring documents the evaluation-order insertion explicitly.


vl3c:
Follow-up commit c5dd928d addresses the four review nits (docstring fold, whitespace, log assertion, three-way interaction test). Answers to the open questions:

**Q1: Has the in-game manual repro been run against this build, or is the unchecked test-plan box still pending?**

Still pending. The fix has been validated via the xUnit tests against the exact StartUT inversion from the user's repro (`b11ef3d4` probe @ 24.26 vs `fa429137` debris @ 20.66) but has not yet been replayed in-game. The unchecked box is honest — the in-game playtest is the user's call to schedule once the PR is otherwise ready.

**Q2: Is there any path where a recording flips `IsDebris` between selector rebuilds within a single re-fly session?**

No. Grep audit of every `IsDebris =` write site in production code:

- `BackgroundRecorder.cs:1187` — `child.IsDebris = !hasController;` at child-recording creation only (decouple event). The `hasController` flag is sampled once at the moment of the split and the resulting `IsDebris` is final for the recording's lifetime.
- `BackgroundRecorder.cs:1304` — same `!hasController` pattern for new vessel infos at creation.
- `BackgroundRecorder.cs:749` — `IsDebris = parentRec.IsDebris` at creation (debris inheritance from parent).
- `ParsekFlight.cs:5911` — `IsDebris = isDebris` at recording creation.
- `Recording.cs:708` — copy constructor (preserves source value).
- `RecordingOptimizer.cs:995` — split helper (preserves original value).
- `RewindInvoker.cs:241` — Rewind provisional inherits from origin (preserves).
- `SessionMerger.cs:147` — merge preserves srcRec value.
- `RecordingTreeRecordCodec.cs:763` — load from sidecar (read-only deserialization).

No write path flips an existing recording's `IsDebris` mid-session based on subsequent controllability changes. The classification is a one-shot stamp at creation by `Coalescer.ProcessBreakupEvent` (the same site that picks "controlled child created" vs "debris child created" log lines). So a cached `PrimaryByPeerInternal` map cannot become stale wrt `IsDebris` within a re-fly session — the only way `IsDebris` changes between rebuilds is if a new recording is created (decouple event), which already triggers a fresh `RebuildFromMarker`.

**Q3: Should Rule 6 stay permanently as a safety net even after the recorder-side parent-anchored surface lands, or get removed once `DebrisParentRecordingId` covers all controlled-decoupled-child cases?**

Agree with the reviewer's suggestion to keep it as a safety net. The recorder-side fix targets re-fly with intra-tree formations (child → parent in the same recording tree); cross-tree co-bubble formations (e.g. two unrelated vessels that happened to share a physics bubble at recording time, like a flyby of an existing ghost) still go through CoBubble peer-blend exclusively. Rule 6 prevents debris from becoming the formation anchor in those cross-tree cases regardless of whether the intra-tree path exists. The Open todo entry already notes this intent ("PR #873 stays as a safety net for cross-tree co-bubble formations") — confirmed.

PR #924, opened 2026-05-20
Plan: GhostPlaybackEngine hot-path decomposition (UpdatePlayback + RenderInRangeGhost)

vl3c:
## Summary

Read-only planning doc (no `.cs` changes) for the behavior-preserving decomposition of the two large per-frame hot-path methods in `GhostPlaybackEngine.cs`: `UpdatePlayback` and `RenderInRangeGhost`. These were flagged as too risky for opportunistic one-shot extraction (many `continue`-terminated guard branches sharing mutable `state`/`ghostActive` locals, allocation-sensitive).

The doc lives at `docs/dev/plans/ghostplayback-hotpath-decomposition.md` and is anchored on M2 already being merged (the `ShouldSkipPostPositionPipeline` + `ResolveRenderSurface` narrow pass).

- Structural map of each method: phases, every guard `continue`/`return`, and which shared locals are read-only vs mutated across blocks.
- 6 ordered, lowest-risk-first extractions (3 per method), each with helper name + signature, the `docs/dev/refactor-guidelines.md` checklist items that make it safe, a zero-new-allocation analysis, and the risk.
- A do-NOT-touch list (guard chains whose extraction would bury a `continue`/`return` or interleave mutations).
- The pure-helper unit tests each extraction enables (new `GhostPlaybackEngineExtractedTests.cs`).
- Allocation hazards and ordering invariants (the #613 relative-anchor retire ordering, the #414 microsecond-rounding invariant).

## Verdict

- `UpdatePlayback`: only partial/conservative decomposition advised. The Phase C finalization tail carves cleanly (`BuildPlaybackBudgetPhases` is the one real testability win); the 15-block per-frame loop body should stay largely inline.
- `RenderInRangeGhost`: moderate decomposition viable (`IsSpawnSuppressedDeadOnArrival` pure predicate + `TryFirstSpawn` + the non-retired post-position block), still conservative.

## Test plan

- [ ] Doc-only PR; no build/test impact. Implementation lands in separate disciplined commits per the ordering in Part 3, with `dotnet test` green between each step and clean-context review gates on the higher-risk extractions.

### DIFF
--- a/docs/dev/plans/ghostplayback-hotpath-decomposition.md
+++ b/docs/dev/plans/ghostplayback-hotpath-decomposition.md
@@ -1,0 +1,669 @@
+# GhostPlaybackEngine Hot-Path Decomposition Plan
+
+Read-only planning doc. NO `.cs` edits land with this commit. This plan covers the two
+large per-frame hot-path methods in `Source/Parsek/GhostPlaybackEngine.cs`:
+
+- `UpdatePlayback(...)` (the per-frame outer loop + diagnostics finalization)
+- `RenderInRangeGhost(...)` (in-range spawn / position / visuals pipeline)
+
+Both were flagged as too risky for opportunistic one-shot extraction: many
+`continue`-terminated guard branches sharing mutable locals (`state`, `ghostActive`),
+and allocation sensitivity (these run for every trajectory every frame, with many
+ghosts active simultaneously).
+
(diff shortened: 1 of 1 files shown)


vl3c:
Superseded by #930, which archives this plan into docs/dev/done/refactor/ with a STATUS header. The decomposition it planned was implemented in #926. Branch left intact.

PR #943, opened 2026-05-21
Record docking-port undock as an Undock branch (fixes crew-ledger corruption)

vl3c:
## Summary
- Docking-port undocks were misrecorded. `OnPartUndock` fires before KSP splits the vessel (the part still reports the old combined persistentId), so the intended Undock branch never ran. The joint-break fallback then classified the departing pod as `PartDestroyed` (docking undocks do not fire `onPartDeCoupleNewVesselComplete`), and the real undocked vessel started ~3.5s late as a Launch branch.
- Subscribe to `GameEvents.onVesselsUndocking` (fires at the end of `Part.Undock()` with both final pids) as the authoritative split signal. `OnPartUndock` now only captures the pre-split structural snapshot and the departing-root-part seed. A new pure `SegmentBoundaryLogic.ClassifyUndockSplit` maps the recorded side; the handler consumes the pending joint-break check to suppress the within-segment fallback, then starts the existing `DeferredUndockBranch`.
- Fixes the downstream crew-ledger corruption: the undocked pod's crew (e.g. Bill Kerman) is no longer recorded as destroyed or flagged permanentlyGone.

## Test plan
- [x] Full xUnit suite green (incl. 7 new `ClassifyUndockSplit` cases + an undocked-pod crew-survival contract test + the in-game handler-wiring contract test)
- [ ] Playtest: dock two crewed vessels, start recording, undock at a docking port. Confirm an Undock branch (not Launch) splits at the undock UT with no PartDestroyed, the undocked vessel records from the undock UT (no ~3.5s gap), and the departing crew stay Aboard. KSP.log should show `onVesselsUndocking: vessel split off ... decision=SplitRecordedStaysActive` and `type=Undock`, and must NOT show `ClassifyJointBreakResult: no new vessels detected ... WithinSegment` for that UT.


vl3c:

Verified the load-bearing KSP contract by decompiling `Part.Undock`: the event ordering (`onPartUndock` -> joint break -> new `Vessel` created -> `onVesselsUndocking` fired last) and the `(oldVessel keeps pid, newVessel gets a fresh pid)` semantics both confirm every assumption the fix depends on. Seed lifecycle, the `restoringActiveTree`/`recorder`/`pendingSplitInProgress`/null guards, the crew end-state path, and the docs are clean. 8 new tests pass.

### Should-fix
- **`SplitRecordedBecomesNew` is unreachable** (`SegmentBoundaryLogic.cs` ~439, `ParsekFlight.cs` ~9912, test ~72). KSP always assigns the split-off vessel a fresh persistentId, so `recordedPid == newPid` can never hold; the branch + its doc-comment + the test assert a state that cannot occur. `DeferredUndockBranch` also hardwires `activeVessel = FlightGlobals.ActiveVessel` and only uses the passed pid as the background vessel, so the computed `activeChildVessel` is never threaded in. Recommend collapsing to `SplitRecordedStaysActive` vs `NotRecordedVessel`, or relabeling the branch as purely defensive and documenting that the coroutine ignores the active-side designation.

### Nit
- **Focus-swap timing** (pre-existing, now first reachable): if the player keeps flying the departing pod, KSP moves focus to the new vessel one frame later (`WaitAndSwitchFocus`), and `DeferredUndockBranch` reads `FlightGlobals.ActiveVessel` after its own one-frame yield, which can resolve to the same vessel as `backgroundPid`. Consider passing both pids and letting the coroutine pick the active side (mirror `DeferredEvaBranch`).
- A genuine within-segment break in the SAME physics frame as the undock would be swallowed by consuming the shared `HasPendingJointBreakCheck` flag (the part is still hidden via its `Decoupled` event; only the `PartDestroyed` SegmentEvent label is lost). Extremely narrow; worth a one-line comment.

PR #944, opened 2026-05-21
Finalize parent-anchored Relative debris off body-fixed surface, not origin-collapsed live altitude

vl3c:
## Summary
- At scene exit, parent-anchored debris with no patched-conic solver built its start state from the live vessel orbit, which returns origin-collapsed coordinates (alt ~= -bodyRadius, e.g. -599034 on Kerbin), tripping the sub-surface check and classifying the recording Destroyed off a bogus altitude.
- The recorded-point contradiction-recovery path was already frame-aware (it reads `bodyFixedFrames` for Relative sections, never the anchor-local metre offsets), but a tight 0.5s window keyed off the bogus live-orbit UT gated it off, so the wrong verdict shipped.
- Extracts `IsRecordedSurfaceContradictionAccepted`: parent-anchored recordings widen acceptance to their own authored coverage span (their live-orbit UT carries no meaning), so the real body-fixed altitude is used. Crashing debris still classifies Destroyed for the right reason (genuine terrain impact); a coasting near-parent controlled child is no longer wrongly classified Destroyed.

## Test plan
- [x] Full xUnit suite green (+5 tests: parent-anchored body-fixed point accepted outside the window and reseeds off the real altitude; crashing debris stays Destroyed; alive near-parent child classified Orbiting not Destroyed; window-decision unit; far-out-of-coverage rejection)


vl3c:

All focus areas cleared. The predicate is gate-only; the surface point is selected nearest-in-time to the real scene-exit UT (`commitUT`), so widening the gate to the recording's authored coverage span cannot pick a wrong (early) altitude. The reseed genuinely uses `bodyFixedFrames` (Relative-frame contract respected, never the anchor-local metre offsets); the parent-EVA structural path still declines safely; crashing debris still classifies Destroyed off its real terrain impact. The two end-to-end tests drive the real finalizer path with a Relative `TrackSection` carrying distinct `frames` vs `bodyFixedFrames` and assert the real altitude is used. 5 new tests pass.

### Nit
- The docstring (`IncompleteBallisticSceneExitFinalizer.cs` ~979) and the CHANGELOG/todo phrase "the live-orbit UT carries no meaning" is imprecise: the UT (`commitUT`) is meaningful; only the live-orbit **position/altitude** is the origin-collapse fingerprint. Reword to "altitude/position."
- `IsRecordedSurfaceContradictionAccepted` keys solely on `ParentAnchorRecordingId != null` (deliberately covering both debris and controlled-decoupled children per the CLAUDE.md orthogonality note). Worth a one-line comment so a future reader does not "tighten" it to debris-only.

PR #945, opened 2026-05-21
Keep RELATIVE anchor engaged when switching to the docking target mid-approach

vl3c:
## Summary
- During a rendezvous, switching control to the target promoted it to the active recording (removing it from BackgroundMap). The still-backgrounded chaser's anchor scan only consulted BackgroundMap and additionally blanket-rejected the active recording, so the chaser silently lost its only RELATIVE anchor and dropped to ABSOLUTE at ~4 m apart for the rest of the close approach.
- New pure `AnchorDetector.TryResolveLoadedAnchorRecordingId` resolves a loaded vessel to its recording via BackgroundMap or the active recording (by pid). The blanket active-recording rejection is removed; cycle safety is preserved by the DAG-order rule (candidate.TreeOrder < focus.TreeOrder, so no A<->B anchor cycle can form), and the re-fly provisional gate is preserved.
- Also downgrades the `TrackSection sparse sampling` WARN to Verbose when warp/on-rails was observed during the section (genuine 1x gaps still WARN), cutting ~50 non-actionable warnings per session.

## Test plan
- [x] Full xUnit suite green (+11 tests: AnchorDetector resolution / DAG-order / cycle-safety; sparse-sampling warp-downgrade decision)
- [ ] Playtest: dock approach, switch control to the target mid-approach, confirm RELATIVE stays engaged (KSP.log keeps `RELATIVE mode`, no `RELATIVE mode exited ... reason=distance-or-missing-anchor` while close).


vl3c:

Cycle safety is airtight: `candidate.TreeOrder < focus.TreeOrder` is a strict comparison over unique, monotonic TreeOrder values, so no A<->B anchor cycle can form (re-fly forks get a higher TreeOrder and can only anchor older recordings); ties resolve to no-anchor. The re-fly provisional marker gate still runs (it was never the removed line), distance + DAG gating remain intact, and the change makes the background path symmetric with the long-standing active-recorder path. Active-recording pid is resolved once per scan, not per vessel. 13 targeted tests pass.

### Should-fix
- **Item 2 per-section warp latch is a real false-negative vs the "1x gaps still WARN" claim** (`FlightRecorder.cs` ~8659, `BackgroundRecorder.cs` ~2169). A single Absolute section can contain BOTH a genuine 1x dropped-sample gap AND a later physics-warp gap (physics warp never goes on-rails, and atmospheric descent does not close the section on warp), so `warpObservedThisSection` latches true and the genuine 1x data-loss WARN is silently downgraded to Verbose. Diagnostic-only impact, not blocking. Either classify per-gap (track which gaps occurred while warp was active), or soften the CHANGELOG/todo wording and note the per-section granularity in the helper doc-comment.

### Nit
- The `bgVessel.packed` term in the BG latch (`BackgroundRecorder.cs` ~2170) is dead: `OnBackgroundPhysicsFrame` early-returns on `bgVessel.packed`, so on-rails samples never reach this tick. Drop the term or fix the "on-rails" comment (the effective BG signal is physics warp only).

PR #951, opened 2026-05-22
Chain-sequential auto looping: consecutive auto-loop chain segments loop as one unit

vl3c:
## Summary

When two or more consecutive primary-path chain segments are all set to Loop with period `auto`, they now play back as a single looping unit: each segment hands off to the next at its boundary and the whole chain wraps seamlessly back to the first segment, instead of each segment relaunching independently on the global per-recording stagger.

Previously every `auto`-looped recording (chain member or not) went into one flat, save-wide launch queue and relaunched on a single fixed gap regardless of duration or chain membership, so consecutive chain segments never synced end-to-start. This adds a shared "span clock" over the unit span `[firstStart, lastEnd]`; the member whose recorded window covers the current span loopUT is the one visible ghost.

Design doc: `docs/dev/design-chain-sequential-auto-loop.md`
Implementation plan: `docs/dev/plan-chain-sequential-auto-loop.md`

## How it works

- **Detection (host-side):** `RecordingStore.DetectChainLoopUnits` groups committed recordings by `ChainId`, finds maximal runs of >= 2 consecutive `ChainBranch == 0` members that are all `LoopPlayback && LoopTimeUnit == Auto` with >= 2 points, and emits opaque index-keyed `LoopUnit` descriptors. No recording-schema change; units are runtime-only.
- **Engine stays chain-agnostic:** `IPlaybackTrajectory` gains no chain fields. The flight engine (`GhostPlaybackEngine`) and the tracking station (`ParsekKSC`) both consume the same `LoopUnitSet` (pushed once per frame) and render members via the shared pure helpers `TryComputeSpanLoopUT` / `TrySelectSpanMember` / `DecideUnitMemberRender`.
- **Span clock:** new pure helper; cadence = span duration (seamless wrap), clamped to `MinCycleDuration`. Overlapping member windows pick the higher `ChainIndex`; inter-member gaps hide all members.
- **Unit members are excluded** from the flat global auto-launch queue in both schedulers.
- **Debris** loop-synced to a unit member follows the span clock (not the parent's own loop clock).
- **Watch mode** retargets the camera at unit-internal boundaries via a new `CameraActionType.UnitHandoffRetarget`, with an owner keep-alive fallback so the watched ghost is never destroyed mid-handoff.
- **Dormant by construction:** with no qualifying chain (empty unit set) every new path falls through to the pre-feature behavior; a zero-`ChainId` fast-out avoids per-frame allocation in that common case.

## Edge cases

All 18 edge cases in the design doc are handled or documented (run-length 1, non-contiguous/manual breakers, branch>0, zero-duration members, overlap, gap, no-spawn-while-looping, anchor gating, warp, etc.). A looping chain spawns nothing (the existing `chain looping` guard), so spawn policy is untouched.

## Test plan

- [x] `dotnet build` clean (0 warnings/errors).
- [x] `dotnet test` green: 12380 passing xUnit (was 12344 pre-feature; +36 new cases in `ChainLoopUnitTests` / `AutoLoopTests` / explainability). The only non-pass is the pre-existing environmental `SyntheticRecordingTests.InjectAllRecordings` skip when KSP holds the log file locked.
- [x] ERS/ELS grep gate (`GrepAuditTests`) passes; serialization no-leak test confirms no new ConfigNode keys.
- [ ] **In-game (requires KSP, Ctrl+Shift+T):** `ChainLoopUnit_SingleGhostAdvancesAndWraps`, `ChainLoopUnit_WatchFollowsLiveMember`, `ChainLoopUnit_DebrisFollowsSpanClock` are written and compile but were not run headless. Confirm in-game: single ghost advances through segments and wraps; watch camera follows the live member (or owner fallback) and never targets a destroyed/inactive ghost; tracking-station map markers replay as a unit (no stray vessel spawn).

## Known v1 limitations (in todo-and-known-bugs.md)

- Watch retarget is single-shot with an owner keep-alive fallback; a dropped transfer at a boundary holds the camera (no retry queue in v1).
- Tracking-station map-marker rendering of units is a confirm-in-game item (Unity-bound, not xUnit-covered).
- No per-chain loop UI toggle yet; units form automatically from per-row Loop + `auto`.


vl3c:
Superseded by the Missions abstraction + whole-mission looping work (#958 and the periodicity/re-aim follow-ups). The span-clock playback mechanism from this PR was reused there; the flag-based chain auto-detection model was replaced by explicit named Mission selections (Mission.ExcludedIntervalKeys / ComputeTrimmedMemberWindows). Closing as obsolete; branch retained for reference.

PR #959, opened 2026-05-24
Keep ghost engine FX and smoke visible out to 5 km, not 2.3 km

vl3c:
## Summary

Ghost engine plumes and smoke were being culled at ~2.3 km in flight (non-Watch mode). A looped launch climbing off the pad lost its exhaust trail far too early.

Root cause: the rendering full-fidelity boundary (`RenderingZone.Physics` -> `Visual`) was hardwired to `DistanceThresholds.PhysicsBubbleMeters` (KSP's 2.3 km physics-load bubble). Crossing it fired two plume-killers at once in `GhostPlaybackLogic.ApplyDistanceLodPolicy`:
- `reduceFidelity` disabled 3 of every 4 renderers, including the engine plume `ParticleSystemRenderer`.
- `suppressVisualFx` stopped engine / RCS / reentry FX and muted audio.

Watched / watch-protected ghosts hit `forceWatchedFullFidelity` and skip both, which is why this only showed in non-Watch flight.

## Change

- New `DistanceThresholds.GhostFlight.FullFidelityRangeMeters = 10000`, deliberately decoupled from the 2.3 km physics bubble.
- The rendering zone boundary (`RenderingZoneManager.FullFidelityRadius`), `ShouldRenderPartEvents`, `LoopFullFidelityMeters`, and the `ApplyDistanceLodPolicy` / `DiagnosticsComputation` reduced-tier thresholds now key off it.
- Hysteresis restore renamed `PhysicsFidelityRestoreMeters` -> `FullFidelityRestoreMeters = 9700`.
- Within 10 km a ghost keeps full mesh + part events + engine / RCS / reentry FX; beyond it drops to the existing reduced-fidelity + FX-suppressed tier.
- The 2.3 km `PhysicsBubbleMeters` still governs relative-frame anchoring and background sampling (unchanged). `ReduceGhostFidelity`, the 50 km mesh-hide, and the 120 km Beyond cull are untouched. KSC scene is a separate code path (warp-only FX suppression + 25 km hard cull) and is not affected.

## Audit

All other distance rules were reviewed and left as-is: 50 km mesh-cull, 120 km Beyond, warp FX-off >10x / mesh-off >50x, KSC FX warp-only + 25 km cull, audio rolloff 30 m - 5 km. Only the 2.3 km full-fidelity boundary was too early.

## Test plan

- [x] Full xUnit suite green (12410 passing). Updated the boundary-coupled `RenderingZoneTests` / `ZoneRenderingTests` / `DistanceThresholdsTests` for 2300 -> 10000 and the renamed constants; added `ApplyDistanceLodPolicy_FiveKmPhysicsZone_KeepsFullFidelityAndFx` plus 5 km loop-spawn assertions that encode the requirement.
- [ ] Live KSP playtest (pure-logic xUnit cannot exercise particle rendering): confirm a non-Watch looped launch keeps its plume / smoke out to ~5-10 km from the camera anchor, and that the mesh still drops to a silhouette past 10 km / hides past 50 km.


vl3c:
Changed to 5 km, not 10 km.

PR #968, opened 2026-05-28
Periodicity: extend to all Kerbol-system bodies (Phase 4)

vl3c:
Completes the deferred Phase 4 of mission periodicity: a looped mission whose target is a cross-parent / interplanetary body now gets a faithful zero-drift relaunch schedule, instead of reading "not aligned" and falling back to the raw launch anchor. Same-parent missions (Kerbin <-> Mun/Minmus) are unchanged.

Built on the merged zero-drift reschedule (#964). Replay-as-is contract: we choose WHEN to launch the recorded trajectory so the live sky matches; we never re-aim or re-plan the transfer. This is NOT a synodic / KAC-style transfer-window model.

Plan: `docs/dev/plans/cross-parent-bodies.md`.

## What works now that didn't

`MissionPeriodicity.ExtractConstraints` walks the body-reference hierarchy to the lowest common ancestor (LCA) of the launch body and each transited target and emits a parent-centric `Orbital` constraint for every body on both downward chains (target side including the target, launch side including the launch body, LCA excluded). The existing multi-constraint zero-drift solver then finds the joint near-coincidence. Every reachable Kerbol body is covered:

- **Direct child** (Mun, Minmus from Kerbin; Gilly from Eve): launch side is empty, only `Orbital(target)` is emitted. Byte-identical to before.
- **Sibling / interplanetary** (Duna, Eve, Moho, Dres, Jool, Eeloo from Kerbin): also emits the launch body's own heliocentric `Orbital(Kerbin)`, so the launch body's phase around the Sun is locked too.
- **Deep** (Ike via Duna; Laythe / Vall / Tylo / Bop / Pol via Jool): adds each intermediate body's parent-centric `Orbital` (e.g. Duna-around-Sun + Ike-around-Duna), deduped by body name.
- **Gravity assist** (Kerbin -> Mun -> Duna): emits same-parent `Orbital(Mun)` + heliocentric `Orbital(Duna)` + `Orbital(Kerbin)`, the launch body emitted once across both chains.
- **Tidally-locked landings** (Mun, Tylo, Bop, Pol): the landing-body rotation flows through the existing Drop/Loose/Tight A/B flag unchanged.

`UnsupportedCrossParent` is retired except for a disconnected body graph (planet-pack pathology); rendezvous stays `UnsupportedRendezvous`.

## How rare the windows really are (and why the look-ahead is large)

Replay-as-is reuses the recorded transfer ellipse (fixed in the Sun's inertial frame), so a faithful launch needs BOTH the launch body and the target back at their recorded ABSOLUTE positions at the same time. That joint coincidence is far rarer than the ~2-year synodic period (which is only the relative geometry, the re-aim quantity): for Kerbin -> Duna it recurs roughly every 1142 Kerbin years (~487k launch-pad rotations). So the schedule's look-ahead horizon is sized to the true joint-coincidence expectation (product of the non-anchor duty reciprocals, clamped to ~2450 Kerbin years), large enough to actually FIND that window for every stock 2-3 body pair. An early draft sized it to the synodic figure and would have stopped ~80x short, silently returning a launch where the target is wherever it happens to be (the transfer flies to empty space); this PR fixes that, and a test pins that Kerbin -> Duna's first launch is a true window where both planets are within their SOI tolerance (the transfer reaches Duna).

## The cap rule for absurdly-long periods

Because faithful interplanetary windows are typically centuries out, the "Time to launch" cell tints amber and shows a years-long countdown when the next relaunch is more than 50 launch-body-years away, to flag how rare it is. The cap is purely display-layer: it never disables the schedule or the "Warp to..." button (which jumps straight to the window), and never downgrades to "not aligned". The threshold is 50 times the launch body's own orbital year (planet-pack safe, with a stock-Kerbin fallback). Genuinely unsolvable configs (rendezvous, disconnected graph) keep the existing "not aligned"; a target whose window is beyond even the generous search horizon (some deep multi-moon chains) falls to a bounded-best amber launch, per the locked "over-constrained configs are never refused" decision.

## No behavior change for same-parent missions

A direct-child config emits the same constraints, the solver picks the same anchor (the pad is both the tightest tolerance band and the shortest period, so the anchor criterion agrees), the look-ahead floors to the legacy value, and the schedule is identical. Verified by the unchanged same-parent tests.

## Also

- `SelectAnchorConstraintIndex` now anchors on the smallest ABSOLUTE tolerance, not the smallest duty cycle: the faithful-window interval is proportional to the anchor's absolute tolerance (the period factors cancel), so the duty criterion would wrongly pin a long-period heliocentric Orbital and make windows millennia-sparse. Byte-identical for same-parent; also resolves the Tylo Tight-mode anchor tie for free.
- A one-shot `CrossParent SOLVED` diagnostic line; the period-cell basis label prefers the target body ("Duna window" not "Kerbin window").
- The Missions-tab loop-unit rebuild is signature-gated (was per-frame), so the larger window search runs once per config change, not every frame.
- No new persisted fields (the schedule is fully derived; `BuildSignature`'s transited-body digest already folds every body the recording transits).

## Tests

New extractor cases (Duna, deep Ike, gravity-assist dedup, Eve -> Gilly, round-trip, disconnected fallback), LCA-walker theory tables, the absolute-tolerance anchor pick, the joint-horizon look-ahead formula, the dominant-label preference, the display-cap helpers, the bounded-best probe-skip branch, a test that Kerbin -> Duna's first launch is a true within-tolerance window (transfer reaches Duna), and an in-game canary (`CrossParentPeriodicityInGameTest`, Kerbin -> Sun -> Duna against the live body graph). Full suite green (12764). Reviewed clean-context per internal phase, plus a whole-PR review and a deep design review that caught and fixed the synodic-vs-absolute look-ahead error.

## Deferred (Phase-4c follow-ups, separately tracked)

The `min(launchSOI/V, targetSOI/V)` launch-heliocentric tolerance refinement, a Tylo Tight-mode anchor-on-launch-body preference, `LookaheadCoverageFactor` / `MaxLookaheadMultiples` tuning for deep planet-pack chains beyond the ~2450-year ceiling, and the analytic (continued-fraction / CRT) joint-resonance solver that would locate the rare window directly instead of brute-force scanning. The synodic / re-aim transfer-window model stays out of scope (locked by the replay-as-is contract).


vl3c:
Superseded by #981 (re-aim interplanetary transfers), now merged to main. Re-aim re-plans the heliocentric transfer to the target's CURRENT position each synodic window, which replaces this branch's replay-as-is Phase 4 approach for the interplanetary case. The body-walk helpers from here were salvaged into the re-aim branch. Closing as superseded.

PR #983, opened 2026-05-31
Re-aim arrival-seam restitch (S4) v1: rotate the recorded arrival onto the re-aimed approach

vl3c:
Closes the re-aim interplanetary ARRIVAL SEAM (the deferred S4 item). v1: the orbit-segment rotation. Implemented autonomously from a twice-reviewed design plan; the rotation math is independently review-verified, but the in-game seam magnitude is NOT yet playtested (no KSP during the build), so this is READY FOR PLAYTEST, not verified in-game.

## The bug
A looped Kerbin to Duna re-aimed mission: the re-aimed transfer reaches Duna's SOI from the re-planned direction, but the recorded Duna arrival hyperbola was spliced on with its ORIGINAL recorded approach orientation. So the ghost reached Duna, jumped back out (~1.37 Gm, about 28 Duna-SOI radii), and re-approached along the recorded direction: the "approached, went back, approached again" you saw after #982 made the descent draw. The seam is pre-existing in the shipped re-aim (#981); #982 only made it visible.

## The fix (v1)
A rigid rotation R about the target body's center maps the recorded incoming approach FRAME onto the re-aimed one, applied to the recorded target-body orbit segments (approach hyperbola + capture + low-orbit descent segments). The approach, capture, and descent now rotate together to follow the re-aimed transfer, so the arrival joins it continuously, down to a small SOI-edge residual (the |v_inf|-magnitude difference, the same accepted class as the departure seam).

Math (corrected across two clean-Opus plan reviews, verified numerically):
- Full 3-DOF FRAME-to-frame rotation aligning BOTH the inbound v_inf asymptote AND the orbital-plane normal (a direction-only match left a residual up to ~1 SOI radius).
- Everything in KSP's swizzled Zup frame (no .xzy in the rotation path; the frame mismatch a first review caught would have corrupted the orientation).
- The inbound v_inf direction is the analytic hyperbolic asymptote from (e, h), not a sampled instantaneous velocity.
- Element read-back via a KSP state-vector round-trip (rotate the Zup state by R, UpdateFromStateVectors) rather than hand-deriving the Zup angle convention: correct by construction.
- Handedness guard: rotate only when dot(h_rec, h_re) > 0; opposite handedness, non-hyperbolic, or degenerate frames fall back to faithful (no rotation) and log.

## Product decision: ROTATE-ALL (landing site moves)
The only viable rigid option. Keeping the recorded landing site fixed while changing the approach direction is NOT a rigid transform (it requires re-solving / inventing an arrival the recording does not contain). So the whole Duna chapter rotates rigidly and the landing site shifts to a re-aimed site consistent with the new approach. This reverses the earlier "keep the landing site" deferral note; the honest alternative was keep-deferred (no rotation). Flagged here for you to confirm or redirect.

## Tests
- Pure xUnit (no Unity): the asymptote, the frame-to-frame rotation (orthonormal, det +1, maps both axes), vector rotation, the time-shift diagnostic. ReaimRotationTests (14) + ReaimSegmentAssemblerTests additions.
- In-game canary (CrossParentReaimCanaryInGameTest): runs the real pipeline through live Orbits and asserts the rotated arrival's v_inf direction + plane normal match the re-aimed approach (dot > 0.999) with ecc preserved.
- Full suite: 13059 passed, 0 failed. Build clean; deployed DLL verified against the build.

## NOT verified in-game (playtest gates merge)
Loop the Kerbin to Duna mission and watch the re-aimed approach on the map:
1. No jump-back-out at the Sun to Duna seam; the orbit line + ghost mesh approach and capture should be continuous.
2. Grep KSP.log for `[ReaimSeam] ... arrival rotated: R-angle=... segsRotated=...` (the one-shot) and the per-frame `ARRIVAL-SEAM Sun->Duna: ... timeShift=...s (computed, not applied in v1)` for the residual.
3. Expect the landing site to have moved (ROTATE-ALL, by design).

## Known v1 limitations (documented fast-follows, NOT bugs)
- The descent-tail POLYLINE still traces the original recorded landing site while the mesh/orbit-line move to the rotated site. The polyline reads raw recorded body-fixed points; rotating it to follow needs a per-frame-cache change best done after a playtest validates R. Separate fast-follow.
- The SOI-crossing TIME-SHIFT is computed and logged but NOT applied in v1 (applying a non-zero shift would re-introduce a smaller seam against the full-span transfer end; ~0 in the validated case). Correct application is a fast-follow.
- A mid-arrival moon flyby (e.g. Ike) would leave the moon-relative segment unrotated (residual at the moon boundaries). Deferred.

Confined to the re-aim path (Source/Parsek/Reaim/ + the resolver + a diagnostic); no change to non-re-aim or same-parent playback. Plan: docs/dev/plans/reaim-arrival-seam-restitch.md.


vl3c:
Closing: superseded by a corrected approach.

The exact-destination requirement makes the ROTATE-ALL approach here wrong. A re-aimed supply-route ghost must arrive at the EXACT recorded destination (a surface base or an orbital station) and replay the recorded arrival + descent / rendezvous VERBATIM, because the recording is relative to that exact destination. Rotating the recorded arrival to fit the re-aimed transfer moves the landing/destination, which is not allowed: the recorded maneuvers must not be deformed.

The correct fix aligns by TIMING at the SOI-entry seam: wait for the window where the transfer naturally arrives at the recorded SOI-entry geometry, with the existing minimal capture-orbit loitering absorbing the residual, so the arrival + descent replay exactly onto the exact destination (the Mun/Minmus phase-lock philosophy applied interplanetary).

Salvaged for the new approach: the pure inbound-asymptote / arrival-v_inf extraction math (reused to MEASURE alignment for the timing search). Dropped: the rotation-and-apply. Branch reaim-arrival-seam is kept for the salvage. New plan + PR to follow.

PR #984, opened 2026-05-31
Re-aim arrival-seam via SOI-entry timing (v1): match the recorded arrival v_inf, never move the destination

vl3c:
Re-aim interplanetary ARRIVAL seam, done the right way: align by TIMING, never by moving the destination. Supersedes the closed PR #983 (which rotated the recorded arrival and was rejected for relocating the landing site). Implemented autonomously from a twice-reviewed design plan; the search method is independently review-verified, but the in-game seam magnitude is NOT yet playtested, so this is READY FOR PLAYTEST, not verified in-game.

## The contract (non-negotiable, from the owner)
A looped supply-route ghost must arrive at the EXACT recorded destination (surface base / orbital station / landing site) and replay the recorded arrival + descent / rendezvous VERBATIM, because the recording is relative to that exact destination. The recorded destination leg is NEVER deformed, rotated, or relocated. Alignment is spent only on transfer timing (and, in a gated follow-up, extending recorded loiter).

## What was actually broken (and what was NOT)
The re-aim already LANDS at the exact recorded spot today: the descent is body-fixed and the re-aim never touches it. The only defect is the cosmetic APPROACH seam: the synthesized transfer arrives at the destination SOI with a v_inf that differs from the recorded approach, so the ghost visibly jumps at the SOI boundary and re-approaches. This PR shrinks that jump without touching the destination.

## What v1 does
Per synodic window, the transfer selection changes from "take the first tof that hits the body's position" to "evaluate the candidate tofs and pick the one whose arrival v_inf best matches the RECORDED arrival v_inf" (the analytic inbound hyperbolic asymptote direction + the sqrt(mu/-a) magnitude, computed apples-to-apples in KSP's Zup frame). Departure stays pad-aligned; the ecliptic projection is kept as a Lambert plane-stabilizer. The recorded arrival/capture/descent stay byte-identical.

## Honest about the limit (measured, not asserted)
Timing can only shrink the seam to an eccentric-orbit floor, not zero it: the synodic window is mean-anomaly-only, so for an eccentric target (Duna e~0.05) the achievable arrival v_inf differs from the recorded by a residual timing cannot remove. v1 therefore LOGS the achieved residual + the seam magnitude and decides per window:
- DECLINE-TO-FAITHFUL when it cannot beat the current seam (chosen seam must be strictly smaller than the faithful seam AND below 0.25*SOI). A declined window is BYTE-IDENTICAL to main today (review-verified), so this can never regress.
- Direct-approach-land eccentric-target missions will commonly decline; that is the expected fail-safe.

## Review-verified make-or-break properties
- Decline-to-faithful reproduces main's exact transfer choice (first-success tof) byte-identically: no silent global regression.
- The entire v_inf path stays in KSP swizzled-Zup (no .xzy slip); candidate and recorded v_inf are differenced apples-to-apples.
- The recorded arrival/capture/descent segments are byte-identical between the timing path and the faithful path (regression test).
- The v_inf math (asymptote, vis-viva sma, eccentricity vector, magnitude) and the seam measurement are correct and fail-closed on degenerate inputs.

## Tests
21 pure xUnit tests (asymptote, magnitude, vis-viva, the objective, the decline truth table, the byte-identical destination regression) + an in-game canary that logs the residual + seam. Full suite: 13064 passed, 0 failed. Build clean; deployed DLL verified.

## NOT verified in-game (playtest gates merge + the follow-up)
Loop Kerbin to Duna and grep KSP.log for `[ReaimPlayback] arrival-timing:` across windows: read `vinfResidual` (m/s), `dirDeg`, and `chosenSeam` vs `faithfulSeam` with `accepted=true|false`. This is the go/no-go number:
- If `chosenSeam` shrinks meaningfully with `accepted=true` and a small residual, the gated loiter-residual follow-up is worth building.
- If the floor stays large, windows read `accepted=false` (decline-to-faithful), which is the expected fail-safe (no regression, landing still exact).

## Gated follow-ups (NOT in this PR)
- Loiter-residual absorption (extend recorded loiter to absorb the arrival-UT residual where a recording has a capture orbit) - gated on the measured residual being acceptable.
- Decline-tolerance tuning from the logged playtest numbers.

Confined to the re-aim path (Source/Parsek/Reaim/ + the resolver + diagnostics); no change to non-re-aim or same-parent playback. Plan: docs/dev/plans/reaim-arrival-seam-timing.md.


vl3c:
Closing: the playtest shows this is a no-op AND fundamentally flawed.

No-op: every window logged accepted=false because the recorded-v_inf extraction returned null (vinfRec=null, scored=0/25), so the objective never fired and every window declined to faithful (= current main). It changed nothing in the playtest.

Flawed even if vinfRec worked: the transfer is rendered over the FIXED recorded span [RecordedDepartureUT, RecordedArrivalUT], so choosing a tof other than the recorded tof (the whole point of this PR) makes the transfer arrive at a different time than the render-span end and overshoot the body. The faithful path already exhibits this (faithfulTof=6820340 != recordedTof=6854613 -> 80 Mm seam) because main nudges the tof when the recorded tof finds no encounter. The timing-objective approach cannot work within the fixed-render-span architecture.

The real problems are in the shipped re-aim (main), not this PR: the transfer overshoots when the tof nudges, and the recorded Duna arrival/descent segments have huge internal seam jumps (653/198/155 Mm between consecutive segments). Needs a rethink, not another patch on this line. Branch kept.

PR #1009, opened 2026-06-01
fix(maprender): body-relative icon-jump metric kills high-warp false positives

vl3c:
## Problem

With `mapRenderTracing` on, the Tier-C `icon-jump` detector fired 20 anomalies on the re-aimed interplanetary ghost "Kerbal X" (rec `61e9177…` member 30), `dPos` running ~4.7x (Sun transfer) to ~56x (Kerbin hyperbolic-escape tail) over the orbit-derived `expectedDP`, clustered at `warpRate=10000`. These were modest (a few Mm), not the old multi-Gm SOI-exit teleports.

## Diagnosis (from `logs/2026-06-01_1838_maptrace-playtest`, not guessed)

Both warp-sensitivity theories are **ruled out** by the data:
- **Non-uniform UT-per-frame:** `currentUT` and the playback `effUT` both advance uniformly at the warp rate (`GhostOrbitIcon` shows `effUT = liveUT − shift` with a constant shift, 1:1, no compression). No UT hitching.
- **Chord-vs-arc on an eccentric orbit:** an ~83 s/frame step is a negligible fraction of these orbits' periods, so the chord ≈ arc ≈ `v·Δt`. Cannot produce >2x, let alone 4.7–56x.

The real cause is a **frame mismatch**: the predicate compared `|Δ Vessel.GetWorldPos3D()|` (raw Unity / floating-origin world frame) against a **reference-body-centered** orbital speed. KSP builds an on-rails vessel's world position as `referenceBody.position + orbitRelative`, so the raw world delta = orbital arc + the reference-body's own world motion. For a ghost far from the floating origin (Kerbal X is 57–77 Mm from Kerbin while `body=Sun`), that reference-body term (plus the 943 Ms loop-shift epoch mismatch on a re-aimed ghost) dominates and scales with geometry, unrelated to orbital speed. The visible scaled-space map icon is orbit-driven and glides smoothly (playtest-confirmed: the heliocentric arc and the ghost on it look fine).

## Fix

`MapRenderProbe` now measures the icon-jump in the orbit's **own reference-body frame** (`bodyRelPos = GetWorldPos3D − referenceBody.position`), so the floating-origin / reference-body world motion cancels and `dPos` is the actual orbital arc that `ComputeExpectedMotionMeters` already predicts. Added a reference-body-change (SOI-crossing) suppression (`bodyChanged`) to the pure `MapRenderTrace.IsIconJump`. Kept the 1 Mm floor (genuine off-orbit teleports still flag), the floating-origin-shift suppression, and the just-reset suppression. The anomaly line now logs `dPos` (body-relative, decision quantity), `expectedDP`, and `dPosWorld` (raw world delta, context).

This kills the confirmed `body=Sun` heliocentric false positives. The `body=Kerbin` in-SOI cases overlap separately-tracked re-aim work; the body-relative metric now reports only a genuine Kerbin-relative divergence there rather than the frame artifact.

## Scope / notes

- Off-by-default dev diagnostic, so no CHANGELOG line; documented in `docs/dev/todo-and-known-bugs.md`.
- **Known limitation (not fixed here):** the detailed-window is UT-based (4–5 s), so at high warp (UT jumps ~83 s/frame) it closes before the next frame and the per-frame `phase=Snapshot` capture does not produce a consecutive-frame sequence under warp. A frame-based window is a follow-up if per-frame warp capture is ever needed.

## Tests

`+2` pure unit tests (`IsIconJump_BodyChanged_Suppressed`, `IsIconJump_SameBody_AboveThreshold_Jump`) plus the new `bodyChanged` arg threaded through existing cases. Full suite green (13172). Build clean; deployed DLL SHA256-verified against the build.


vl3c:
Clean-context Opus review (spawned directly, not `/ultrareview`): **SHIP AS-IS**, no must-fix / no should-address. It independently verified the central claim against the decompiled `Assembly-CSharp`: `OrbitDriver.updateFromParameters` positions on-rails vessels as `SetPosition(referenceBody.position + pos - localCoMoffset)`, and `GetWorldPos3D()` / `CelestialBody.position` are in the same world frame (FloatingOrigin shifts both equally), so `worldPos - body.position` cancels the contamination exactly. Edge cases (genuine-teleport detection preserved, `bodyChanged`/SOI state machine, `justReset` redundancy, `dPosWorld` log) all confirmed sound.

Addressed its three doc nits in 87807af4 (comment/doc only, no behavior change). One correction to the **Diagnosis** section above: the frame contamination is purely the **reference body's own world motion under warp** — the "943 Ms loop-shift epoch mismatch" is a red herring. `worldPos` and `body.position` are read at the same live UT, so the loop shift only determines where on its orbit the ghost sits, not the contamination.

PR #1031, opened 2026-06-04
Smooth sparse body-fixed map polyline legs via render-time subdivision

vl3c:
## The bug

In map view, the ghost trajectory line for the phase right before a landing (the deorbit / atmospheric-entry arc above a body) rendered as straight lines stitched into a visible polygon, while the dense in-atmo descent below it was smooth. They meet at a UT boundary, so there was a visible smooth-to-faceted step.

## Cause analysis (confirmed and quantified from KSP.log, save s15)

The faceted phase is recording `61e9177193444e329247d0e8288cf91e`, leg 15:

```
DRAWN-leg15=[70963373.6,70963652.6] span=279s body=Duna pts=39 alt0=58245 altN=49998 lon0=73.9 lonN=108.6
```

It is a body-fixed polyline leg with only 39 recorded samples across ~35 degrees of Duna longitude (~0.91 deg/segment). Per-segment geometry for every drawn leg of this recording:

| leg | body   | pts | dLon (deg) | deg/seg | chord km | sagitta m |
|-----|--------|-----|------------|---------|----------|-----------|
| 0   | Kerbin | 14  | 0.50       | 0.038   | 0.4      | 0.0       |
| 1   | Kerbin | 53  | 1.60       | 0.031   | 0.4      | 0.0       |
| 2   | Kerbin | 84  | 31.50      | 0.380   | 4.8      | 4.0       |
| 8   | Kerbin | 96  | 24.60      | 0.259   | 3.4      | 1.9       |
| 13  | Duna   | 50  | 5.70       | 0.116   | 1.2      | 0.3       |
| 14  | Duna   | 27  | 3.60       | 0.138   | 0.9      | 0.3       |
| 15  | Duna   | 39  | 34.70      | 0.913   | 6.0      | 11.9      |

Leg 15 is the coarsest by ~2.4x (0.913 deg/segment, ~12 m sagitta), and crosses 34.7 degrees in only 38 segments, so it reads as a visible 38-sided polygon at map zoom. The dense descent below it (`0b91670cf8334780a2b78687e80d6923`, ~200 samples) is smooth. The dominant artifact is within-leg chording, which subdivision fixes (not a junction tangent kink: leg 15 ends at the same alt 49998 / UT boundary where the dense descent begins).

Two structural facts (left untouched):
- It is NOT the conic anchor. The log shows `Anchor leg SKIPPED (one-sided): ... reason=one-sided-bracket`; `TryAnchorLegToConicSeam` only Slerps points and cannot introduce facets.
- It is a body-fixed polyline (not a smooth orbit conic) because FIX #27 (`IsOrbitSegmentBelowSurface` / `ComputeOrbitalCoverIntervals`) excludes the below-surface deorbit OrbitSegment from the orbital cover and hands its samples to the polyline (`excluded 2 below-surface orbit segments from cover ... -> descent leg [70963373.6,70963652.6]`).

## The fix (render-side only, recorded data immutable)

New pure helper `GhostTrajectoryPolylineRenderer.DensifyBodyFixedArcs` (`internal static`, no Unity native calls) inserts interpolated vertices between consecutive recorded samples so each drawn chord spans at most `DensifyTargetSegmentDegrees` (0.3 deg) of great-circle arc, hard-capped at `MaxDensifiedPointsPerLeg` (512).

- Interpolation Slerps the body-fixed surface unit normal and linearly interpolates altitude + UT, so inserted points land on the body (correct `GetWorldSurfacePosition` mapping) and it is robust at the +/-180 longitude seam and near the poles.
- Recorded samples are preserved exactly as vertices; only points BETWEEN them are inserted. No `.prec` / OrbitSegment / recorded-point / recorder / load path is modified.
- Densification runs ONCE at leg-build time inside `BuildLegFromBodyFixedPoints`. The Driver's per-frame loop iterates `leg.PointCount` over the larger arrays unchanged, so there is no new per-frame cost beyond the extra vertices (respecting the many-ghosts efficiency principle).
- A leg already finer than the target (the 200-point dense descent) is returned verbatim, so it does not regress or balloon. The hard cap bounds pathologically coarse legs by scaling per-chord insertion counts down proportionally.
- One-shot `Polyline densify:` build summary (Verbose, batch-counted) logs pre/post point counts, the subdivided-leg count, and the target resolution.

The other coarse-but-high-altitude legs (leg 2 at 0.38 deg/seg, leg 8 at 0.26) are also smoothed by the same target; no residual follow-up is needed (no junction discontinuity dominates this case).

## Validation

- `cd Source/Parsek && dotnet build`: clean (0 warnings, 0 errors).
- `cd Source/Parsek.Tests && dotnet test`: 13320 pass. 19 new tests for the densify helper (endpoints + interior recorded vertices preserved, UT strictly increasing, sparse arc subdivided to target, already-dense returned unchanged, hard cap respected, +/-180 seam and pole robustness, altitude lerp, single-point passthrough, great-circle / Slerp helper math, end-to-end through the builder). The only non-pass is `InjectAllRecordings`, a dynamic-skip guard that refuses to run because a live KSP holds the shared install's KSP.log (unrelated to this change).

## Docs

- `CHANGELOG.md`: one user-facing line under 0.10.0 > Fixed.
- `docs/dev/todo-and-known-bugs.md`: new Done entry with the cause analysis and fix summary.


vl3c:
Closing: investigation showed this densify change does not fix the reported near-landing faceting. That artifact is the conic-to-body-fixed deorbit seam under the re-aim rotation mismatch (the arrival hold lands the deorbit at a per-loop-varying rotation phase, measured 17 km to 224 km off), not sparse-sample chording, which is sub-pixel at planet scale here. The real cause is being investigated separately. Branch left in place.

PR #1104, opened 2026-06-09
Test isolation: full single-test coverage, TRACKSTATION baseline, and campaign-save safety

vl3c:
## Summary

Makes the in-game test runner safe to use against a real campaign: running any test or batch now leaves the player's save untouched, every `AllowBatchExecution = false` test can run isolated, and a stuck building or a cancelled run always returns you to a usable Space Center.

## What's in here

**1. All "single-run" tests are now isolatable.** Audited every `AllowBatchExecution = false` test; 34 already carried the FLIGHT quicksave-baseline isolation. The two remaining holdouts are fixed below (scene-exit merge dialogs promoted earlier in the branch; the Tracking Station "Fly" canary generalized below).

**2. Facility force-close safety net.** The Phase-5 `StockUiOverlay` tests enter a facility building (which pauses the game) and close it in a `finally` — but a user **Cancel** stops the coroutine with Unity `StopCoroutine`, which skips that `finally`, leaving you stuck inside a paused building with no recovery. Added `ForceCloseOpenSpaceCenterFacilities` (gated to SPACECENTER) that fires the stock despawn `GameEvent` — exactly what UISpaceCenter's exit button does — for any open R&D / Astronaut Complex / Mission Control. Wired into between-runs cleanup, `Cancel()`, and per-test cleanup. (Administration is deliberately excluded — see the note below.)

**3. TRACKSTATION baseline isolation.** The baseline capture/restore was FLIGHT-only. Generalized it to FLIGHT **or** TRACKSTATION so the last manual-only Tracking Station "Fly" canary runs under **Run All + Isolated** (launched from the Tracking Station) and returns there afterward. The FLIGHT restore-core scene commit is byte-identical; only the scene-specific steps branch on a new `CapturedScene`. The non-FLIGHT commit mirrors the one proven in-repo non-flight scene resume (`TriggerSaveAndExitToSpaceCenter`: `HighLogic.CurrentGame = game` → `Game.Updated()` → `SaveGame` → `HighLogic.LoadScene`).

**4. Campaign-save safety (defensive audit).** Enumerated every disk write the test framework makes under `saves/<save>/` and closed the gaps:
- **`persistent.sfs` was unprotected** — `TriggerSaveAndExitToSpaceCenter` writes it directly (the two scene-exit tests now in the isolated batch) and scene-change auto-saves rewrite it; the baseline restore only reloaded the baseline *slot* into memory. The batch now backs up `persistent.sfs` (to a non-`.sfs` `<slot>-persistent.bak`) and restores it byte-for-byte via the atomic `FileIOUtils.SafeWriteBytes` path at teardown (batch end **and** cancel).
- **Player's stock `quicksave` slot was clobbered** by two tests using the default slot → now unique throwaway slots deleted in `finally`.
- **`parsek_ledger_groundtruth.sfs`** left behind → deleted in the harness `finally`.
- **`.loadmeta` sidecars** now swept by `TryDeleteSaveSlot`.

All new file I/O is defensive: try/catch, atomic safe-write, and **preserve-the-backup-on-failure** so a failure degrades to a warning rather than corrupting the campaign.

## Review

One adversarial Opus review was run on the full diff. Findings addressed:
- **HIGH** — non-FLIGHT resume diverged from the repo's proven `Game.Updated()` pattern → rewritten to mirror `TriggerSaveAndExitToSpaceCenter`.
- **LOW** — `FindObjectOfType<Administration>` could match the strategy-lifecycle test's hidden (disabled, `Instance`-set, `activeInHierarchy`) Administration canvas → Administration branch dropped (no test opens it as a real building, so this removes the only false-positive path with no loss).
- **LOW** — doc wording clarified (FLIGHT capture/cleanup also gained the persistent backup/restore).

New pure helpers are unit-tested in `InGameTestRunnerTests` (scene gating, availability reason, facility-close summary).

## Validation status

Built and playtested on a real career (2026-06-09, logs `logs/2026-06-09_2319_pr1104-test-isolation/`):

- **xUnit:** 14,791 passed; the only failure is the known environmental `InjectAllRecordings` flake (KSP was running and holding `KSP.log`).
- **In-game:** full `Run All + Isolated` from all four scenes: EDITOR 89/92, SPACECENTER 113/124, TRACKSTATION 108/112, FLIGHT 281/340 (rest skipped), 4 failures triaged below.
- **Isolation lifecycle verified in KSP.log:** 4 batch captures (DiskOnly for SPACECENTER/EDITOR, InMemoryAndDisk for TRACKSTATION/FLIGHT), 4 `persistent.sfs` backups + 4 restores, 4 crash-marker write/clear cycles, the TRACKSTATION baseline restore after the TS canary, 36 per-test FLIGHT restores, facility force-close fired twice (Mission Control), and crash-reconcile correctly skipping with `same-process-no-crash`.

The 4 in-game failures were craft/save-specific and are fixed in the follow-up commit (`c7237c0e3`):

- **3x `WaitForRecordingToLeavePrelaunch` timeout (10s):** the clamp release shipped in this PR fired correctly, but the career craft's first stage (4x Shrimp SRBs) has marginal TWR and was still creeping off the pad at the deadline with engines burning (the sibling auto-record test cleared PRELAUNCH at 8.5s of its 10s window). Both launch waits now use a 30s deadline and route their timeout through the pure `ClassifyLaunchWaitTimeout`: still-PRELAUNCH with thrust produced is a Skip (craft performance, environmental), no-thrust keeps the existing Skip, and a timeout after leaving PRELAUNCH stays a Fail (recording-side regression).
- **1x `RecordingStopMetricsValid` REC-002:** a committed single-point EVA recording in the career save ('Lars Kerman': 1 point, no terminal-state stamp, splashed `TerminalPosition`, sealed on board-back) fell outside the terminal-Landed-only exemption. Broadened to `IsNeverLeftGroundSinglePointRecording` (terminal Landed, or no terminal stamp plus a recorded surface `TerminalPosition`).
- 13 new xUnit tests cover both helpers (`LaunchWaitTimeoutClassifierTests`, `NeverLeftGroundExemptionTests`).

Remaining caveats:

- The Tracking Station "Fly" canary itself skipped with "SpaceTracking selection/Fly reflection helpers are unavailable": a pre-existing reflection probe (`BtnOnClick_FlySelectedVessel`) that does not match this KSP build, so the canary content has not run; the TS baseline capture/restore around it did.
- Crash reconcile was only exercised on its same-process skip path; an interrupted-batch (force-kill) playtest is still pending.
- A re-run of `Run All + Isolated` on the rebuilt DLL is needed to confirm the four fixes in-game.


vl3c:
Follow-up: merged latest `main` and addressed the review findings; an independent clean-context re-review of the follow-up commit (`18ae5b4e4`) confirms it correct.

**Fix:** `CommitNonFlightSceneLoad` no longer uses `Game.Updated()` + `SaveGame(slot)`. Decompiling `Assembly-CSharp` (1.12.5) confirmed `Game.Updated()` does not round-trip its receiver: in a planetarium scene it rebuilds `flightState` from the live scene and pulls `scenarios` from the global `ScenarioRunner`, then returns `this`. So the old code published the post-test session, and `SaveGame(refreshed, slot)` overwrote the pristine baseline slot, corrupting it for subsequent restores in the batch. It now replays stock KSP's own in-game load sequence (the `PauseMenu` Load button): `HighLogic.CurrentGame = game` -> `GamePersistence.UpdateScenarioModules` -> set `startScene` -> `Game.Start()` (which also runs `CrewRoster.Init` + `Status = ONGOING` that a bare `LoadScene` skipped). The baseline slot is now read-only during restore.

This supersedes the "non-FLIGHT resume ... rewritten to mirror TriggerSaveAndExitToSpaceCenter" note in the description above.

Also corrected the stale `todo-and-known-bugs.md` bullet to match the shipped `Start()` sequence, and documented the facility-open `FindObjectOfType` detector's audited invariant.

`dotnet test`: 14738 passed, 0 failed, 0 skipped. In-game playtest still pending (Tracking Station "Run All + Isolated" is the path to watch).

PR #1107, opened 2026-06-09
Mission / Route structure-list window (0.10.1)

vl3c:
## Summary

Adds the **Structure window**: a popup that lists a run step by step (launch, staging / separation, dock, undock, deliveries, terminal / landing), each with its time and location. It complements the Missions-tab composition-over-time tree with a flat "what happened, in order" view, and is the readable alternative the roadmap chose over map route lines.

This is the first Tier-1 item from the post-0.10.0 logistics remaining-work path (`docs/roadmap.md`, Phase 13). It is additive and read-only over already-recorded data: no capture, scheduler, serialization, or route-model change.

One reusable window covers both:
- a **mission** (Missions-tab "Structure" button on each mission header), and
- a **supply route** (Logistics-window "Structure" button in each route's detail panel).

## How it works

Two pure, headless-tested builders feed a reusable IMGUI sub-window (mirrors the `SpawnControlUI` / `CareerStateWindowUI` ownership pattern, owned by `ParsekUI`, drawn in KSC + Flight):

- `MissionStructureListBuilder` flattens the cached `MissionStructure` (root launches + branch-point dock / undock / decouple / eva / breakup events + terminal legs) and supplements it with debris-staging `PartEvents` (`Decoupled` / `FairingJettisoned` / `ShroudJettisoned`). A `Decoupled` event is de-duplicated against a controlled-decouple branch point by **decoupler part PID** (not UT, which differs across recorder paths); fairing / shroud pass through. Steps are sorted into a total chronological order.
- `RouteStructureListBuilder` emits Origin then Dock then Delivery (per stop) then Undock, from `Route.Origin` / `IsKscOrigin` / `RecordedDockUT` / `Stops` plus the dock-member recording's `RouteConnectionWindow`. The source recording is resolved through an injected lookup so the builder stays free of the store singleton.

Per-step location is the coarse body / launch-site / terminal context (reusing `RecordingsTableFormatters`); per-step coordinate resolution from trajectory points is deferred.

## Tests

- `StructureListBuilderTests` (12 xUnit cases): launch+terminal, controlled-decouple dedup with debris/fairing pass-through, cross-recording staging dedup, dock-before-undock ordering, EVA, determinism, verbose-summary log line, route KSC-origin ordered steps with manifest summary, vessel origin, missing-source, null route, location formatter.
- `StructureListRuntimeTests` (in-game): builds the lists against live committed trees / routes and asserts non-empty + UT-ordered.
- Full suite green (14,778 passing).

## Notes

[...]

PR #1109, opened 2026-06-09
Fix post-couple dock-side transport baseline (false MixedPickupDelivery reject)

vl3c:
## Summary

Fixes a logistics route-creation bug: a clean cargo-delivery Supply Run could be wrongly rejected as `MixedPickupDelivery` when the transport and the destination shared a fuel type that stock crossfeed equalised during docking. Closes the open known limitation "Logistics route window dock-side baseline is post-couple".

## Root cause

`RouteAnalysisEngine.HasResourcePickup` rejects a run when `transportGain = UndockTransportResources - DockTransportResources > epsilon` (the transport appears to have picked cargo back up). `DockTransportResources` was extracted from the **post-couple merged vessel snapshot**, which `CreateMergeBranch` captures several frames after the couple (via the deferred dock-confirm path). A same-frame stock crossfeed equalisation that drained the transport tank into the depot therefore deflated the dock baseline, and the later undock reading then read as a spurious `+transportGain` on a resource the player never intended to move.

The **endpoint side** of this was already fixed (`endpointPreCoupleSnapshot` / `pendingDockPartnerSnapshot`); the transport side had no equivalent.

## Fix

Mirrors the proven endpoint mechanism for the transport side:

- `ParsekFlight.OnPartCouple` now captures the recorder's own (transport) pre-couple vessel snapshot into `pendingDockSelfSnapshot` / `pendingDockSelfSnapshotPid`, in the same block where `data.from` / `data.to` are still distinct (pre-reparent), via a new `ResolveSelfVessel` helper that matches `recorder.RecordingVesselId` regardless of survivor status. Captured in both the main and retroactive blocks; cleared at every site the partner snapshot is cleared.
- `RouteProofCapture.BuildDockRouteConnectionWindow` takes a new `transportPreCoupleSnapshot` parameter and extracts `DockTransportResources` / `DockTransportInventory` from it, selected **self-validatingly** (`SnapshotContainsAnyPartPersistentId(snapshot, transportPids)`, else falls back to the merged snapshot — current behaviour preserved when absent or PID-mismatched).

The heuristic alternative (approximate-equalisation detection in `HasResourcePickup`) was rejected: it could misclassify a genuine net-out pickup.

## Scope / caveat

Route-creation eligibility only: no change to dispatch, delivery, serialization, or the route data model. The headless tests prove the **selection logic** flips the classification; whether stock crossfeed equalisation lands after the `onPartCouple` event (vs partly during `Part.Couple`) is the same playtest-confirmable assumption the shipped endpoint fix relies on. The merged snapshot is captured frames later regardless, so the fix is a strict improvement. **In-game playtest pending.**

## Tests

`RouteProofCaptureTests` (5 new): transport baseline pre-couple vs merged, no-override fallback, self-validation fallback on mismatched PIDs, inventory parity, and an end-to-end `AnalyzeRecording` repro that is rejected as `MixedPickupDelivery` with the deflated baseline and accepted as `Eligible` (a clean 100 LF delivery) with the pre-couple baseline. Full suite green (14,729).

## Notes

- Version bumped to 0.11.0 (`AssemblyInfo.cs`, `KSPAssembly` tag, `Parsek.version`). Coordinates with the structure-window PR #1107, which also targets 0.11.0; whichever lands second resolves a trivial same-version CHANGELOG/AssemblyInfo overlap.
- Plan: `docs/dev/plan-dock-side-baseline-fix.md`.


vl3c:
Updated per maintainer: this ships under the existing **0.10.1** section (no version bump). Reverted the AssemblyInfo / KSPAssembly / Parsek.version bumps; the CHANGELOG entry now lives under `## 0.10.1`'s Bug Fixes. (Supersedes the "bumped to 0.11.0" note in the description.)

PR #1137, opened 2026-06-13
Map-view trajectory render: forward-window SOI bridge fix (Bug 2) + zoom-cull and burn-seam icon instrumentation (Bug 1, Bug 3)

vl3c:
## Summary

Three map-view trajectory render issues surfaced in the "orbital supply route" save (looped Mun supply
missions watched in map view, including at high time warp). All three are pre-existing behaviors in main,
exposed because these missions loop and phase-lock at warp. Root cause was traced from a live KSP.log with
the map render tracer enabled.

Read the per-bug status before reviewing: only Bug 2 is a complete behavioral fix. Bug 1 ships a
provisional fix plus diagnostics, and Bug 3 ships diagnostics only, both gated on one in-game playtest.

## Bug 2 (FULL FIX) - approach line not drawn ahead of the icon across an SOI

The forward render window hard-stopped at the first SOI change, so a looped Kerbin-to-Mun (or any
moon/planet) transfer drew the destination-approach line only after the ghost icon crossed into the
destination SOI, never ahead of it.

`ForwardRenderWindow.ComputeForwardWindow`'s forward walk now spans exactly one SOI transition into the
destination body (bounded by an explicit `soiTransitionsConsumed` counter capped at 1). It stops at the
destination's first full-loop closed orbit after the crossing (new reason `DestinationClosedOrbit`, the
capture/parking ellipse), a same-SOI closed orbit before any crossing (existing `FullLoopClosedOrbit`), a
second SOI change (new reason `SecondSoiBound`, so an interplanetary Kerbin-Sun-Duna walk stops at the
Sun-Duna seam rather than drawing the whole heliocentric transfer), or end-of-data. The backward walk is
untouched, so the destination-side run still resets at the seam and the Mun ghost never redraws the Kerbin
ascent behind it. The `IconOnClosedOrbit` reset is preserved.

## Bug 1 (PROVISIONAL FIX + instrumentation) - looped-mission lines vanish when the map is zoomed out

Looped missions are dominated by short near-surface body-fixed legs (ascent/descent) whose Vectrosity
polyline meshes frustum-cull when the map camera pulls back; stock orbit lines (large-extent meshes) do
not. After every `VectorLine.Draw3D()` at the three draw sites the fix stamps an effectively-infinite
mesh-bounds box (re-applied each frame, because Vectrosity recalculates the AABB synchronously per draw).

This is provisional: it is correct only if the vanish is a frustum cull. An instrument-first probe (gated
on `mapRenderTracing`) logs `PlanetariumCamera.Distance`, the leg AABB, `GeometryUtility.TestPlanesAABB`,
the post-draw bounds, and a one-shot MeshFilter component-path check. If the frustum test stays true at the
vanish zoom, the cause is sub-pixel collapse instead, the override is a no-op, and the real fix is a
distance-scaled minimum width floor. Confirm in-game before relying on the override.

## Bug 3 (INSTRUMENTATION ONLY, behavioral fix deferred) - followed icon teleports one frame at a burn

The user saw the followed icon teleport for one frame near a burn. The log shows the headline event fires
with the orbit element unchanged (only the position jumped), right after the state-vector reseed cleared
the ghost's segment-drive bounds: the segment-to-state-vector (burn) seam suppresses the proto icon for a
couple of stale frames, then it snaps when the next drive re-establishes. (This is one instance of the
broader warp-reseed-lag icon class, 12 icon-off-orbit + 7 icon-teleport anomalies across 8 ghosts.)

The behavioral continuity fix depends on a one-frame ordering read that a headless run cannot perform, and
mutating the no-bounds suppress path risks regressing the kept-permanent no-conic / below-atmosphere /
off-arc Director fallback. So this commit ships diagnostics only: a rate-limited "Icon suppress
{Enter|Exit}" Verbose line carrying frame, liveUT, last-suppressed frame, and the proto world + body-
relative position, so one playtest log shows the stale-window length and the snap magnitude. No user-facing
behavior changes.

## Tests

- `ForwardRenderWindowTests`: 5 existing tests rewritten for the single-SOI-span contract + 5 new
  (Kerbin-to-Mun capture spans seam, icon-on-last-departure-element, flyby-no-capture bounds at second SOI,
  interplanetary bounds at second SOI, destination-side backward-walk regression guard).
- `GhostOrbitIconDriveTests`: pure suppress decision both ways, the four transition cases, gap-frame
  Exit-then-Enter, stale-stamp guard, dict reset.
- `GhostTrajectoryPolylineMeshBoundsTests`: pure AABB compute + the infinite-bounds constant; the
  Vectrosity-coupled write and far-zoom frustum pass are covered by an in-game test.
- Full xUnit suite green (15263 passed; `InjectAllRecordings` is the known shared-install environmental
  flake, non-blocking).

## Validation done on this branch

Each bug was adversarially validated, plan-reviewed before coding, and code-reviewed after. The review pass
caught and fixed a real defect: the Bug 3 instrumentation was initially wired into the wrong suppress
branch (it would have emitted zero lines for the documented repro) and was moved to cover the
Director-traced path the headline event actually takes, with a regression test added.

## Remaining (after an in-game playtest of "orbital supply route", mapRenderTracing on)

- Bug 1: read the zoom-cull probe at the vanish zoom to confirm frustum cull (keep the override) vs
  sub-pixel collapse (replace with a width floor).
- Bug 3: capture the "Icon suppress" Enter/Exit lines to implement the behavioral continuity fix.
- Bug 2: confirm the approach line is continuous from SOI entry to the destination orbit before the icon
  crosses (D21 04:36 to 04:38; re-check D24 03:43).


vl3c:
## Bug 3 playtest data arrived (and it splits the fix into two facets)

The playtest this branch was waiting on is captured in `logs/2026-06-13_1608_m4c-recording-deepdive/`
(a higher-warp re-capture of the same looped Mun/Kerbin ghost family, at 100000x). I verified its
headline claims against the raw `KSP.log`:

- Anomalies: `icon-off-orbit` **537** (vs 12 in the prior 100-1000x capture), `line-blink` 62,
  `polyline-orbit-overlap` 5, `icon-teleport` 2 (`warpRate=100000`). `directorDrive=True` on all 54 evals.
- 107 "loop-shifted member, deferring to loop-aware per-frame create" lines.
- Smoking-gun single-frame lifecycle (`KSP.log:16606/16636/16637/16652`): ghostPid 21256233 is **created,
  its FirstPosition equals the spawn position, flagged `icon-off-orbit angleIconVsOrbitEff=46.51
  angleEffVsLive=0.00 loopShift=0.0`, and destroyed - all in frame 34817**. For recording ed43b6fb alone:
  339 create-segment-done = 339 destroys.

### Why this matters for the deferred Bug 3 fix

There are **two facets**, and the originally-planned fix only covers one:

- **Facet (a) - persistent-proto reseed-lag** (lower warp, the original 100-1000x capture): the applied
  orbit/icon position goes stale until the ~0.5 s reseed timer fires. Covered by the planned reseed-seam
  fix (re-drive the icon position on `mapReseedHeadLeftSegment`).
- **Facet (b) - per-frame destroy+recreate** (extreme warp, this capture): loop-shifted protos are fully
  rebuilt every frame, so the icon is read at end-of-frame before stock `OrbitDriver` settles, then the
  proto is destroyed. `angleIconVsOrbitEff` is uniformly random 0-180deg (read-before-settle signature),
  while `fromOrbit==toOrbit` / `angleEffVsLive=0` confirm the orbit math is correct - only the rendered
  position lags. **There is no persistent proto for the reseed seam to freshen**, so facet (a)'s fix does
  not cover this. The per-frame-create path must reposition-not-recreate, OR drive the fresh proto's icon
  before the end-of-frame probe reads it. This also collapses the 62 `line-blink` anomalies and removes the
  per-frame `ProtoVessel.Load/unload` churn.

Full write-up (two-facet design, loci, guards, hard dead-ends, success metric) is in
`logs/2026-06-13_1349_map-trajectory-render-bugs/ROOT-CAUSE-ANALYSIS.md` (Bug 3 section, updated) and the
handoff brief `logs/2026-06-13_1608_m4c-recording-deepdive/HANDOFF-fix-map-trajectory-render.md`.

Build caveat: that log came from the M4c DLL (base-main anomaly detection, PR #1014), not this branch, so
it lacks this branch's burn-seam instrument (bdc76282e). The behavioral evidence above is from base-main
instrumentation and is directly usable; deploy this branch to capture the burn-seam instrument output for
facet (a). Bug 1 still needs its own in-game `TestPlanesAABB` read (no anomaly class, not in this log).

vl3c:
## Bug 3 facet (b) fix pushed (604e74131); facet (a) was rejected as a no-op

Pushed `604e74131` "Settle looped ghost map icons before the probe reads them (Bug 3 facet b)".

### What shipped (facet b - drive-before-probe)
At extreme warp, loop-shifted map ghost ProtoVessels are destroyed+recreated every frame (the per-frame
destroy is forced by `GhostPlaybackEngine` - the loop unit member falls outside its ~785 s window when a
100000x frame advances ~4000 s). The create-time icon drive (`ForceImmediateIconDrive`) runs before
`ShadowRenderDriver.RunFrame` writes the new pid's StockConic seed, so the Prefix sees `fresh=False`, takes
the legacy effUT path (discarded by KSP's live re-propagation), and the end-of-frame `MapRenderProbe`
records the icon at its raw-epoch spawn position - the read-before-settle `icon-off-orbit` family.

Fix: the flight per-frame create queues each freshly-created loop-shifted pid;
`GhostMapPresence.SettleFreshlyCreatedLoopGhostIcons()` (called from
`ParsekFlight.UpdateTimelinePlaybackViaEngine` AFTER `RunFrame` writes the seed and BEFORE the
order-10000 probe) re-drives those icons via a second `updateFromParameters()`. The seed is now fresh, so
the Prefix's Director branch runs `SeedAndDriveLive` (bakes the loop shift into the orbit epoch via an
absolute `SetOrbit`), and the settled on-orbit position survives KSP's live re-propagation. Empty-set fast
path (zero cost on frames with no loop create); reuses the existing effUT->world resolution (no new shift
math, no body-fixed->inertial lift); does not touch the suppress path; flight-scoped (the shared create is
untouched, so no Tracking-Station leak). Unit-tested; full xUnit suite 15269 pass.

### What was deliberately NOT done (facet a rejected)
The originally-planned facet (a) (an eager same-frame icon re-drive in the reseed seam) was found to be a
**guaranteed no-op** for the documented population: its planned site is the `loopEpochShiftSeconds == 0.0`
skip-unchanged branch of `RunFlightMapOrbitReseedPass`, which the anomaly population never traverses, and
`docs/dev/todo-and-known-bugs.md` had already rejected that approach + the warp-scale-timer idea. Re-pinning
the moderate-warp 1349 population showed 8 of its 12 `icon-off-orbit` anomalies are `fresh=False`
(Director seed absent -> legacy effUT path discarded) - i.e. the **same drive-before-probe mechanism** as
facet (b), at lower warp. So the facet-(a) agent committed nothing and recommended folding the
moderate-warp population into the facet-(b) fix, which the create-time settle is expected to cover.

### Validation (human, both warp regimes)
Deploy this branch and playtest "orbital supply route" with `mapRenderTracing` on at MODERATE warp AND
100000x; `icon-off-orbit` + `icon-teleport` should drop to ~0. Tripwire: a settle-summary line logs
`settled/directorActive/sampleWorldPos`; `directorActive=0` with `settled>0` means the frame ordering
regressed. If residual moderate-warp anomalies remain, that is the persistent-proto sub-case for a
follow-up (the reposition-persistent-proto path was rejected here as a lifecycle rework). A
Tracking-Station-following warp playtest would tell whether to mirror the settle pass there too. Bug 1
still needs its own in-game `TestPlanesAABB` read (untouched).

vl3c:
Closing in favor of #1143. The Bug 1 (zoom-out cull) and Bug 3 (warp burn-seam icon teleport) fixes were moved verbatim onto a clean branch off `origin/main` (#1143). The Bug 2 work here (extending the forward render window across the SOI seam) was the wrong approach: it violated the render-only-within-the-current-SOI invariant. The real Bug 2 is a WITHIN-SOI forward gap (the line is not continuous from the SOI entry to the destination orbit while the icon is inside the SOI); it is now an OPEN item in docs/dev/todo-and-known-bugs.md and needs an in-game screenshot at D21 ~04:36 to localize the exact missing piece before a fix. Branch `fix-map-trajectory-render` is left intact for reference.

PR #1143, opened 2026-06-13
Map-view ghost trajectory: zoom-out cull fix (Bug 1) + warp burn-seam icon teleport fix (Bug 3)

vl3c:
## Summary

Two independent looped-ghost map-render fixes, off a clean `origin/main`. This replaces
[#1137](https://github.com/vl3c/Parsek/pull/1137): that branch also carried a "Bug 2" change that extended
the forward render window across the SOI seam, which was the wrong approach (it violated the
render-only-within-the-current-SOI invariant) and has been dropped. The real Bug 2 (a within-SOI forward
gap from the SOI entry to the destination orbit) is recorded as an OPEN item in
`docs/dev/todo-and-known-bugs.md`, pending in-game localization, and is not part of this PR.

## Bug 1 - looped-mission trajectory lines vanish when the map is zoomed out (provisional + instrument-first)

The tiny near-surface body-fixed Vectrosity leg meshes frustum-cull when the map camera pulls back; stock
orbit lines (large-extent meshes) do not. After every `VectorLine.Draw3D()` at the three draw sites
(`TryDrawLeg`, `TryDrawSeamBridge`, `DrawForwardArc`) the shared `OverrideLineMeshBoundsAfterDraw` helper
stamps an effectively-infinite mesh-bounds box (re-applied each frame because Vectrosity's
`RecalculateBounds` clobbers it).

Provisional: correct only if the vanish is a frustum cull. A gated `Bug1 zoom-cull probe` Verbose line logs
`PlanetariumCamera.Distance`, the leg AABB, and `GeometryUtility.TestPlanesAABB` so the mechanism can be
confirmed in-game (`frustumPass` goes false at the vanish zoom -> keep the override; stays true -> revert
and use a minimum on-screen width floor instead).

## Bug 3 - followed icon teleports / goes off-line for one frame at a burn under warp (drive-before-probe)

Loop-shifted map ProtoVessels are destroyed+recreated every frame (the engine retires a loop unit member
once it falls outside its window, which at warp is every frame), so the create-time icon drive runs before
the shadow writes this frame's StockConic seed; the freshly-built icon is read at its raw spawn position
before settling, then snaps when the next segment establishes a drive. `SettleFreshlyCreatedLoopGhostIcons()`
(called from `ParsekFlight` after `ShadowRenderDriver.RunFrame` writes the seed and before the
order-10000 `MapRenderProbe`) re-drives the freshly-created loop pids so the Director branch's
`SeedAndDriveLive` lands the settled on-orbit position before the probe reads it. Empty-set fast path
(zero cost when no loop ghost was created), reuses the existing effUT->world resolution, does not touch the
suppress path, and is flight-scoped (the shared create site is untouched, so no Tracking-Station leak).

## Tests

- `GhostTrajectoryPolylineMeshBoundsTests` (Bug 1): pure AABB compute + the infinite-bounds constant; the
  Vectrosity write + far-zoom frustum pass via the `GhostTrajectoryPolyline_MeshBoundsOverride_SurvivesFarZoomFrustum`
  in-game test.
- `GhostOrbitIconDriveTests` (Bug 3): the pure settle decision, the suppress-transition classification, and
  the set-clear-on-reset / empty-set fast path.
- Full xUnit suite green (15319 passed; `InjectAllRecordings` is the known shared-install environmental
  flake, excluded).

## Remaining (in-game, human)

- Bug 1: read the zoom-cull probe at the vanish zoom to confirm frustum cull (keep the override) vs
  sub-pixel collapse (switch to a width floor).
- Bug 3: warp 100-1000x following a looped ghost and confirm the icon no longer teleports/goes off-line at
  the burn seam (grep `Icon suppress` Enter/Exit; the `icon-off-orbit` / `icon-teleport` anomalies should
  drop to ~0 with the tracer on). Validate at both moderate warp and 100000x.
- Bug 2 (separate, OPEN): the within-SOI forward gap still needs an in-game screenshot at D21 ~04:36 to
  localize the exact missing piece before a fix.


vl3c:
Closing and splitting into two focused PRs off origin/main: #1144 (Bug 1, zoom-out cull - provisional + instrument) and #1145 (Bug 3, warp icon teleport - fix). Same code, cleanly separated; the combined branch map-render-zoomcull-iconteleport is left for reference.

PR #1149, opened 2026-06-14
Fix loop ghost map icon off its orbit line at warp (state-vector seam CoMD lag)

vl3c:
## Summary

At high time warp, a looped transfer ghost's map icon flicked far off its own orbit line (up to ~168 deg, reading as a brief **duplicate/displaced icon** near the destination) while coasting through an orbit-raise burn. Found while validating PR #1136. This is the residual after PR #1145 (which fixed only the creation-frame placement).

## Root cause (decompile-confirmed, KSP 1.12.5)

The surviving anomalies are all the **state-vector gap-glide** path, not the director drive: in the capture every large `icon-off-orbit` carries `loopShift=0.0` + `reason=orbit-raise-gap-points`, while director-drive frames stay < 3 deg.

- `Vessel.GetWorldPos3D()` (what the icon renders at) returns the cached `Vessel.CoMD`. It self-recomputes only while `!precalc.firstStatsRunComplete` (true only at creation - which is exactly why #1145's creation bake works but a steady-state re-drive does not).
- `CoMD` is a `VesselPrecalculate.CalculatePhysicsStats` (FixedUpdate) product; for a packed orbiting ghost it is `mainBody.position + orbitDriver.pos`.
- `OrbitDriver.updateFromParameters(bool)` re-propagates `orbit.pos` to now and `SetPosition`s the **transform**, but never writes `CoMD`.
- The gap-glide reseed (`GhostMapPresence.UpdateGhostOrbitFromStateVectors`) runs in a Parsek per-frame pass **after** FixedUpdate, so it changes the conic while `CoMD` still holds the pre-reseed phase for one frame. At warp the `(currentUT - reseed-epoch)` gap leaves the icon up to ~168 deg off its line, snapping back next FixedUpdate. The director path is clean because its reseed runs **inside** FixedUpdate alongside `CalculatePhysicsStats`.

A second `updateFromParameters()` (an earlier proposal) would be a no-op: it never touches `CoMD`.

## Fix

After the reseed, force `vessel.precalc.CalculatePhysicsStats()` (the same refresh `GetWorldPos3D` itself uses) so the icon lands on its conic the same frame. Gated to the conic-change seam by the pure `GhostOrbitElementsChanged` (sma/ecc/epoch, mirrors `GhostOrbitBodyChanged`) so steady-state reseeds add no per-ghost cost. `ghostLastAppliedOrbitElements` is cleared at all 6 `ghostLastAppliedOrbitBody` teardown sites. Covers both flight and tracking-station via the shared method.

## Observability (no-op trap guard)

Shared-key proof-of-fire: `Seam CoMD re-snap summary: refreshed=N steady=M maxOffDegPostRefresh=X`. Acceptance on a warp re-fly playtest = `refreshed > 0` and, with `mapRenderTracing` on, `maxOffDegPostRefresh < 1`, plus zero subject `icon-off-orbit` > 10.

## Tests

- **xUnit** `LoopIconWarpLagTests`: the `GhostOrbitElementsChanged` gate (sma/ecc/epoch incl. hyperbolic Mun-capture, NaN/Inf, within-epsilon) and the `IconVsOrbitAngleDeg` proof metric.
- **In-game** `LoopBoundaryIconWarpLagTest` (`GhostMap`, FLIGHT): proves at runtime the reseed leaves `CoMD` stale (> 20 deg off) and `CalculatePhysicsStats` re-snaps it (< 5 deg) - refuting the substituted-second-drive no-op trap - plus the steady-state skip.
- Full headless suite green (15398).

## Review

Read-only adversarial review (3 lenses): SHIP-WITH-NITS, zero blockers; nits applied (guard the summary `string.Format` behind `IsVerboseEnabled`, runtime steady-state assertion). Packed-not-landed branch reachability, PhysicsFramePatch packed early-return, the segment-drive-clear contract, and all teardown sites verified intact.

## Validation status

Headless + in-game tests written; end-to-end warp behavior needs a manual playtest (the acceptance greps above). Tracking-station coverage is structurally identical (shared method) but was flight-only in the capture.


vl3c:
Merged latest `main` (resolved a docs conflict on the Bug 3 todo entry that #1151 also touched).

**Scope note after the merge:** this PR fixes the **state-vector gap-glide (orbit-raise) CoMD seam** specifically (the `loopShift=0.0` + `reason=orbit-raise-gap-points` frames from `logs/2026-06-14_1335_kerbalx-terminal-dup-icon/`). The merge pulled in #1151's stage-B note documenting a separate `icon-teleport` burst (176 anomalies at 10000-100000x on a **hyperbolic Kerbin-escape** ghost `pid=4156666876`, `fromOrbit==toOrbit`). That is a different ghost population and that note attributes it to the rate-limited 0.5 s map-orbit reseed cadence (the icon riding a stale **segment** orbit between reseeds), not the per-FixedUpdate CoMD seam fixed here, so it is likely **not** covered by this PR and is flagged as still-open in the todo doc. Confirming which path that ghost takes (state-vector update vs Director/segment reseed) needs its own capture.

vl3c:
## Validated in-game (2026-06-14 warp playtest)

Both CoMD re-snaps fire and the systemic warp icon lag is resolved on both paths:

| | before fix | after both fixes |
|---|---|---|
| `icon-off-orbit` | 26 (up to 158 deg) | **8** |
| proof-of-fire | — | `Seam ... refreshed=6`, `Director icon CoMD re-snap summary: refreshed=5683` |

The persistent main-ghost offenders from the pre-extension run (`pid=496223269` @152 deg, `pid=217146972` @158 deg) are **gone**.

**The 8 survivors are a separate, smaller facet** (not what this PR targets): almost all are the **first frame after a ghost is created** (anomaly frame = GhostCreated + 1) on short-lived ghosts that exist <1 s, several on parent-anchored "Kerbal X Probe" sub-ghosts that render via the body-frame/grace path - a sibling of the #1145 creation-frame facet, not the steady per-frame CoMD lag the two re-snaps fix. Tracked as a follow-up. Logs: `logs/2026-06-14_1729_loop-icon-warp-validate2/`.

Ready to merge.

PR #1153, opened 2026-06-14
Fix ghost transfer trajectory filling in progressively at an SOI seam

vl3c:
## Symptom

When a Kerbin-to-Mun transfer **ghost** crosses into Mun's SOI on the flight-scene map, only a short segment renders at the crossing (the captured-ellipse forward arc + a small seam bridge + the recorded tail leg). The incoming curve toward Mun fills in progressively, as the ghost travels, instead of appearing all at once.

## Root cause

At the SOI/segment seam the incoming hyperbola is owned by neither render surface:

- **Stock ghost orbit line** (which owns the conic the icon currently rides) gets blanked by the stale-segment guard (`GhostOrbitLinePatch`, `reason=stale-segment-awaiting-reseed`) the moment the playback head passes the applied single-segment upper bound.
- The **grace debounce** that would bridge that reseed lag was disqualified for hyperbolas: `ShouldDeferOrbitLineHide` returned false for any non-elliptical orbit (`IsOrbitFiniteElliptical` requires ecc < 1), and the incoming conic is a hyperbola (ecc ~1.75).

The per-frame map reseed applies one `OrbitSegment` per tick (two consecutive same-sma hyperbola segments, then the capture ellipse), and the early head-left reseed only runs at `TimeWarp.CurrentRate > 1`, so the conic revealed segment-by-segment as the head walked it.

## Fix (Lever B: un-blank the stock line that already owns the conic)

`ShouldDeferOrbitLineHide` gained an optional `orbitFiniteHyperbolic` flag. For the `stale-segment-awaiting-reseed` reason the renderable-arc gate now accepts a finite ellipse **or** a finite hyperbola (`GhostOrbitArcPatch` already draws a valid open hyperbolic arc), so the line is held visible across the reseed lag instead of blanked.

- New pure `IsOrbitFiniteHyperbolic` (ecc > 1, finite; exactly-parabolic excluded to match the arc clip's `> 1.0` gate), computed once in the Postfix and passed only to the stale-segment call.
- The `polyline-owns-phase` grace stays **ellipse-only** (flag defaults false there), so a hyperbolic dip never defers the line over a polyline-owned phase: **no double-draw**.
- Boundedness: the hold reuses the same frame-count grace window (`OrbitLineGraceFrames` = 20), last stamped on a genuinely-shown frame and never restamped while hidden, so a hyperbolic hold expires within ~20 render frames (~0.33s @ 60Hz) of the last genuine show. Consecutive same-sma hyperbola segments share the same conic, so holding the prior one during the lag is visually identical; the only stale window is the single hyperbola-to-ellipse capture change, capped by the frame bound.
- The grace-defer log now records `arc=hyperbola|ellipse` so the SOI-seam hold is greppable apart from the ordinary elliptical reseed-lag defer.

This is **Surface B** of the two-surface seam gap. Surface A (the `GhostTrajectoryPolylineRenderer` forward-arc current-element exclusion) is a separate lever **not** taken here: the incoming hyperbola is the stock line's responsibility by design, so forcing the forward-arc pass to also draw it would double-draw the same conic on the reseed-catch-up frame.

## Tests and validation

- 7 new pure cases in `GhostOrbitLineGraceTests` (stale-segment + hyperbolic inside grace defers; after-grace hides; polyline-owns + hyperbolic still hides; neither-shape hides; elliptical regression unchanged; durable + hyperbolic never deferred).
- Full headless suite green: 15568 passed, 0 failed.
- In-game playtest pending (load-bearing): loop a Kerbin-to-Mun transfer, warp through the SOI crossing, and grep `KSP.log` for `reason=stale-segment-awaiting-reseed arc=hyperbola` to confirm the path fires; verify `polyline-orbit-overlap` / `line-blink` anomaly counts do not increase.

## Notes

- **Overlap with #1149** (`fix-loop-icon-warp-lag`, open): a different bug (one-frame icon-off-line CoMD lag). The only shared file is `GhostOrbitLinePatch.cs`, in a disjoint region (this change touches `ShouldDeferOrbitLineHide` / `IsOrbitFiniteHyperbolic` / the stale-segment Postfix branch; #1149 touches the icon-CoMD path). `GhostMapPresence.cs` is read-only here. Mechanical conflict only; whichever lands second resolves a trivial hunk.
- **Out of scope**: at 1x the early head-left reseed is gated to `TimeWarp.CurrentRate > 1`, so a brief re-blank is possible if the grace expires before the next 0.5s reseed timer. The repro is warp-only; the durable fix is a separate eager-reseed follow-up.


vl3c:
Closing: in-game playtest (logs/2026-06-14_2358_mun-soi-fix-check) shows this fix is misdirected. The progressive-reveal symptom is dominated by director-traced-path-suppress + terminal-below-atmosphere suppression of the stock orbit line (off 56 of 63 frames during the crossing) and forward-arc under-coverage (Surface A), NOT stale-segment-awaiting-reseed, which is a single one-off transition recovered the next frame by the body-frame layer in both the original and the post-fix logs. Reopening investigation on the correct surface. Branch left in place.

PR #1156, opened 2026-06-14
Fix looping transfer ghost orbit line drawing one SOI-approach segment at a time

vl3c:
## Symptom

A looping Kerbin-to-Mun transfer **ghost** on the flight-scene map draws only the **first** piece of its incoming Mun trajectory at the SOI crossing; the rest of the curve toward Mun appears ~27 s later (it "fills in" as the ghost travels rather than all at once).

## Root cause (proven by the arc-clip logs)

The stock ghost orbit line is **on and drawing** the whole crossing (`reason=visible-body-frame lineActive=True`), but its drawn **arc sweep is geometrically clipped to one OrbitSegment at a time**. The Mun approach is stored as 2-3 adjacent OrbitSegments with **identical** Kepler elements (the recorder closes + reopens a segment at every recording-mode transition - this split is by design). The clip log is the smoking gun:

- `Arc clip ... arc=11.8deg ... drawEnd=179 pts=180` (one segment), then ~27 s later `arc=112.0deg` (the next segment, contiguous in eccentric anomaly: `toE` of piece 1 == `fromE` of piece 2). `drawEnd=179 pts=180` is a true geometric clip, not a point-budget cap.

For a **loop-shifted** ghost, `GhostMapPresence.TryGetVisibleOrbitBoundsForGhostVessel` short-circuits to the single stored applied-segment bounds (`ghostOrbitBounds[pid]`, written one segment per tick) and **skips the equivalent-segment merge** the non-loop path already runs (`TrajectoryMath.TryGetOrbitWindowForMapDisplay` / `ExpandEquivalentOrbitWindow`). So the arc clips to one same-orbit fragment; the next is revealed only when the playback head advances into it.

The line on/off gate uses a **separate** wide body-frame window (`ghostBodyFrameOrbitBounds`) and keeps the line on correctly, so this is an arc-sweep bug, not a visibility/suppression one. (An earlier attempt that targeted the stale-segment line-blank, PR #1153, was closed: an in-game playtest showed it had no effect, because the line is already on.)

## Fix (Option A: expand-on-read; recording stays immutable)

- New pure `TrajectoryMath.TryExpandStoredSingleSegmentWindow` locates the raw recorded segment matching the stored window and grows it across element-equivalent adjacent neighbours (`AreOrbitSegmentsEquivalentForMapDisplay` + an `isPredicted`-match guard mirroring `CoalesceSameOrbitFragments`).
- `GhostMapPresence.ExpandStoredBoundsAcrossEquivalentSegments` un-shifts the stored live-frame bounds to raw (`MapLiveUTToEffUT`), expands on the raw committed `OrbitSegments`, then re-applies the loop shift, writing back to the arc-clip window only when more than one fragment coalesces. Applied to the loop-shifted **and** endpoint-tail short-circuit branches.
- **Shape-clip only:** `ghostOrbitBounds` (which drives `SetOrbit` + the icon epoch) is never mutated, so the icon stays on the single applied segment and only the drawn arc widens. The recorder / optimizer are unchanged.
- **Stops before a different conic:** the element-equivalence predicate folds the adjacent hyperbola fragments and stops before the captured ellipse (`sma=551701`), so it never draws a wrong arc.

This is the non-re-aim, loop-shifted follow-up that the 2026-06-02 Playtest-2 fix (`ReaimSegmentAssembler.CoalesceSameOrbitFragments`) explicitly anticipated; that one merged the same same-orbit fragments but was scoped to re-aim members only.

## Tests

- 6 pure cases in `OrbitSegmentTests` (`TryExpandStoredSingleSegmentWindow_*`): merges adjacent equivalent and stops at the non-equivalent boundary; single-segment / no-seed-match / predicted-boundary / null-empty; seed-by-startUT tolerance.
- 1 wrapper case in `GhostMapEndpointTailTests` driving `ExpandStoredBoundsAcrossEquivalentSegments` end to end: asserts the un-shifted-expanded-then-re-shifted window (stopping before the ellipse) and the `Loop arc-window coalesced ... fragments=2 segIndices=0-1` log line.
- Full headless suite green: 15569 passed, 0 failed.

## Validation status

**In-game playtest is PENDING and load-bearing - not confirmed fixed on green xUnit alone.** To validate: loop a Kerbin-to-Mun transfer, open the flight map, warp through the SOI crossing, and confirm the incoming Mun approach draws as one continuous line (no ~27 s reveal lag); grep `KSP.log` for a single continuous `Arc clip` window instead of the `arc=11.8deg` then `arc=112.0deg` two-piece reveal, and confirm the icon position and the captured-ellipse boundary are unchanged.


vl3c:
In-game validated on the merged-fix DLL (logs/2026-06-15_1835_mun-soi-fix2-check, mapRenderTracing on).

- Coalesce fires on the Mun approach: `Loop arc-window coalesced reason=loop-shifted fragments=2 segIndices=3-4` (merges the two equivalent hyperbola segments, stops before the non-equivalent captured ellipse seg5).
- The Arc clip now draws a single continuous `arc=123.9deg` on every occurrence across both loop members, instead of the prior `arc=11.8deg` then `arc=112.0deg` two-piece reveal (both old values now absent = 0).
- No regressions: 0 Parsek errors, 0 `icon-off-orbit` on the Mun ghosts, over-merge guard held. The few `icon-teleport`/`line-blink` anomalies are the separate known warp-reseed icon facet, not this arc-clip-window change.

Merged `origin/main` (incl. #1149) cleanly; full suite green at 15592.

PR #1166, opened 2026-06-15
Engage re-aim on a heliocentric-parking (two-burn) interplanetary departure

vl3c:
Engages re-aim for a looped interplanetary mission whose recorded transfer departs from a HELIOCENTRIC PARKING orbit (a two-burn departure: escape into a near-circular co-orbital solar orbit, coast, then burn for the target), and re-phases that park into the live frame so the departure renders correctly. Save s15 "Kerbal X #2" (Kerbin->Duna).

Now contains two landed increments:

1. Classifier engage (`ReaimClassifier.IsHeliocentricParkingDeparture`): re-aim recognizes a two-burn departure instead of declining it. Admissible = a closed solar loiter of exactly `DefaultKeepRevs` whole revs (empty cut), near-circular, co-orbital with the launch body; sub-period MCC / multi-rev / eccentric / wide-sma parks fail closed to faithful. Adds the `DepartedFromHeliocentricPark` plan flag + per-member ReaimDiag logging.

2. Increment 1, park re-phasing (#1167, merged into this branch): `ReaimPlaybackResolver.BuildWindowSegments` rotates the verbatim park's LAN by `Delta_lon = omega_parent * (D_k - RecordedDepartureUT)` so the Sun-inertial park renders near live Kerbin instead of ~239 deg off, connected to the escape and the re-aimed transfer. Gated on `DepartedFromHeliocentricPark` + a near-equatorial guard; fails closed; byte-identical when off.

Both gated so every non-heliocentric-park mission (Duna One direct, Mun/Minmus, eccentric targets, station loops) is byte-identical. Recorded data is never mutated (render-time rotation only).

## Validation
- Headless: full suite green (15,702; InjectAllRecordings env-flake excluded).
- 3-agent adversarial review on the classifier engage + fixes; clean-context design review on the park re-phasing.
- In-game (2026-06-16): re-aim engages, the park renders near live Kerbin (239 deg teleport gone), the FIRST synodic window renders fully correct, and the transfer reaches Duna.

## Deferred follow-ups (tracked, not regressions; gated out of every other mission)
- Increment 2 (self-overlap scheduling): a heliocentric-park mission's recorded span (~270 d) exceeds the Kerbin->Duna synodic (~227 d), so the single-instance loop relaunches on the span cadence and the transfer drifts ~20 deg per window off live Duna for windows after the first. The fix is to relaunch every synodic with self-overlap (the previous instance still in flight). Design: `docs/dev/plans/reaim-self-overlap-scheduling-design.md`. Window 0 is correct today.
- The center-to-center cross-SOI seam is the separate #1165 work.

Do not modify PR #1155 (P4).


vl3c:
Held (converted to draft). The classifier-engage fix here is correct and tested, but it engages re-aim for a heliocentric-park (two-burn) departure whose recorded span exceeds the Kerbin->Duna synodic period, which the current re-aim scheduler cannot render correctly (it relaunches on the span cadence, so the re-aimed transfer lands ~142 deg off live Kerbin for every window after the first, the launch is not pad-aligned, and the verbatim park renders ~239 deg off). Every real heliocentric-park interplanetary mission is span>synodic, so this engage has no correctly-renderable case on its own.

The fix is a self-overlap re-aim scheduling milestone (relaunch every synodic with self-overlap, resolver re-keyed per window, synodic pad-align, and the park re-phased into the live frame). This PR will ship together with / be superseded by that milestone. Do not merge alone.

PR #1168, opened 2026-06-16
Fix looped-ghost launch-leg to escape-conic seam jump in map and tracking station

vl3c:
## Problem

A looped ghost's body-fixed launch leg was drawn at the live planet rotation while its escape conic is inertial with the loop shift baked into the orbit epoch. The two diverged by the body spin accrued over the loop shift (82 to 146 degrees measured on the s15 / Kerbal X #2 mission). The 45 degree seam bridge then skipped, and the map / tracking-station icon visibly jumped from the end of the launch line onto a far point of the escape hyperbola.

Diagnosed from `logs/2026-06-16_1821_s15-ts-icon-seam`: zero Kerbin icon-teleports and 243 `nominal` per-frame drive samples (the per-frame icon drive is healthy in Kerbin SOI), with 6 `Seam bridge skipped (angle too large) ... angleDeg=85/82/146/143/139 max=45.0` lines on rec `aa48920e`. So this is a discrete seam-handoff bug, not a per-frame drive bug.

## Fix

Counter-rotate each one-sided launch-leg point's longitude through the existing pure helper `BodyFixedLongitudeAtUT` (per-point `recordedUTs`, not `headUT`) into the conic's inertial frame, at both:

- the draw site (`TryDrawLeg` point build), and
- the decide site (`DecideSeamBridges` leg-side endpoint). This second site is load-bearing: without it the decide pass still measures the full gap and skips the bridge, making the draw fix inert.

The residual seam collapses to a few degrees and the existing 45 degree bridge arms unchanged, drawing one continuous line from the ascent into the escape conic. The launch line stays fully visible.

## Scope and guardrails

- Gated on a loop-shifted ghost (`GetGhostOrbitEpochShift != 0`) AND a one-sided leg (`!IsRunLegAnchorCandidate`). Non-looped / live ascents pass raw longitudes verbatim (a structural, byte-identical no-op) so an in-progress ascent stays glued to the rotating pad. Two-sided burn-leg anchor legs are left to `TryAnchorLegToConicSeam`.
- The 45 degree cap, the seam-bridge builder/drawer, the anchor path, and the conic LAN are untouched.
- `liveUT` is threaded through `PendingLegDraw`, `TryDrawLeg`, `TracedPathTreatment.TryDrawOwnedLeg`, and `DecideSeamBridges`.
- The correction also covers one-sided descent trails of looped ghosts (same divergence), which is consistent with the run-leg-to-inertial-conic contract.

## Tests

- xUnit (`GhostTrajectoryPolylineBuildTests`): the helper collapses an 82 to 146 degree seam below the bridge gate; a small residual still arms; `shift == 0` leaves longitudes byte-identical; a log-assertion confirms the skip wording disappears once corrected.
- In-game (`RuntimeTests`, GhostMap category): FLIGHT and TRACKSTATION seam-collapses-to-conic-frame, plus a non-looped-ascent-stays-glued-to-pad regression.
- Full headless suite green (15732).

## Validation

Headless green and deployed locally. In-game visual and runtime-test validation in the flight map and tracking station is the remaining step.


vl3c:
Closing: the launch-leg-into-inertial-frame approach is the wrong fix (it detaches the ascent from the launch pad and bifurcates the trajectory). The looped launch->escape seam is the re-aim launch-pad-alignment facet in the span >= synodic regime: re-aim engages but ReaimWindowPlanner.PadAlignLaunch opts out because cadence != synodic, so the launch is never rotation-aligned. The real fix extends PadAlignLaunch to that regime; tracked on a fresh branch.

PR #1170, opened 2026-06-16
Re-aim cross-SOI fix P2-P4: patched-conic chain synthesis (escape+transfer+capture) behind the default-OFF flag

vl3c:
Geometry phases (P2-P4) of the looped re-aim cross-SOI transfer fix, stacked on the merged P0/P1 foundation (#1169). **Default-OFF flag (`reaimChainSynthesis`); flag-OFF is byte-identical to today.**

## Status: regressed in playtest, then fixed + hardened (NOT yet validated in-game)
A flag-ON playtest (Kerbal X #2) rendered **worse** than baseline — the synth escape/capture legs were garbage hyperbolae (ecc 8-13: position sampled at the SOI boundary paired with the Lambert *endpoint* velocity at the planet center). A deep review found that plus 3 more issues. All fixed in the latest commit:

- **Fail-closed leg-sanity gate (the safety net):** new pure `ReaimChainGeometry.IsSaneLegConic` (ecc in [1,3]; periapsis `a(1-e)` in [body surface/atmo, 0.5·SOI]). A leg that fails it → `null` → the assembler keeps the recorded leg verbatim / the chain falls back to `ReplaceHeliocentricLeg`. **The captured ecc-12.9/7.77 garbage both fail the gate, so flag-ON can never render worse than baseline.**
- **Root cause:** `BuildBodyRelativeLeg` now samples velocity via `transfer.getOrbitalVelocityAtUT(crossingUT)` (self-consistent with the position at the same UT), not the endpoint `v1`/`v2`.
- **Capture bisection** walks backward from arrival to a valid inside/outside straddle (proximity path had `soiEntryUT==arrivalUT`).
- **Per-leg phase pin** lands each leg's SOI-crossing on the recorded-span boundary.
- **Flag-OFF is now a true no-op** (leg math + `chain legs:` log gated on the flag).

## What P2-P4 build
- **P2** — capture `v1`/`v2`, synthesize escape/capture legs (pure `ReaimChainGeometry` + SOI bisection), plan spans + `CaptureSynthesizable` (Ike-thread/atmospheric → false).
- **P3** — `ReplaceTransferChain` (contiguous tiling, capture pinned, capture fail-closed on Ike, all-or-nothing fallback), `useChain=true`; STEP 4 re-verified vs merged #1155.
- **P4** — `AssertSaneWindowSegments` drops `Count==3` and now asserts escape AND capture legs are sane hyperbolae; real Ike-threaded E2E fixture + in-game chain tests.

## Validation
Build clean; **15798 passed, 0 failed** (`InjectAllRecordings` excluded as the KSP-lock flake), incl. 9 new `IsSaneLegConic` tests.

**Not yet validated in-game.** The live escape/capture geometry can't be unit-tested headless; it needs a **clean direct (no parking-loiter, no moon-encounter) Kerbin→Duna** looped recording — neither Duna One (threads Ike) nor Kerbal X #2 (heliocentric-parking departure) is a clean case, so both fail closed on at least one side. Until that validation, the fail-closed gate guarantees worst-case == the current baseline. **Do not merge before the clean-mission playtest.**

## Deferred (out of scope, fail closed correctly)
Heliocentric-parking departures and the Ike-threaded (multi-moon) arrival; the center-to-center Lambert premise. Minor: the `UniformShift…` test comment is stale vs the per-leg pin (test still valid).

### DIFF
--- a/Source/Parsek.Tests/ReaimChainGeometryTests.cs
+++ b/Source/Parsek.Tests/ReaimChainGeometryTests.cs
@@ -1,0 +1,183 @@
+using System;
+using Parsek.Reaim;
+using Xunit;
+
+namespace Parsek.Tests
+{
+    // Pure state-vector geometry for the re-aim whole-chain synthesis (reaim-fix-plan.md P2, STEP 1/2):
+    // the body-relative velocity / position reductions, the .xzy round-trip identity (a swizzle error
+    // silently corrupts the synthesized orbit), and the SOI-sphere-crossing bisection. The live glue
+    // (Orbit.UpdateFromStateVectors, CelestialBody.orbit, PatchedConics) stays in ReaimTransferSynthesizer
+    // and is validated by the in-game canary; these pure pieces are the xUnit floor.
+    public class ReaimChainGeometryTests
+    {
+        // ---- vRel / rRel reductions (the frame-agnostic subtraction the construction feeds UpdateFromStateVectors) ----
+
+        [Fact]
+        public void RelativeVelocity_SubtractsBodyVelocity()
+        {
+            // vRel = v_heliocentric - v_body. Both operands in the same frame -> a plain subtraction.
+            var vHelio = new Vector3d(12000.0, -3000.0, 500.0);
+            var vBody = new Vector3d(9000.0, -2500.0, 100.0);
+            Vector3d vRel = ReaimChainGeometry.RelativeVelocity(vHelio, vBody);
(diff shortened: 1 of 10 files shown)


vl3c:
Closing: the chain-synthesis approach does not solve the re-aim arrival-render bug it targeted, and regressed the flag-ON render on the available test missions (all of which are hard cases it fails closed on). Reverting the P0/P1 foundation (#1169) too. Baseline stands.

PR #1174, opened 2026-06-17
Re-aim: close the looped launch->escape seam (borrow-repay + boundary overlap)

vl3c:
## Problem

A looped re-aim mission whose recorded span exceeds its Kerbin->destination synodic period (`cadence == span > synodic`, e.g. mission s15 / Kerbal X #2, recording aa48920e) rendered the body-fixed launch ascent rotated 82-220 deg off the frozen inertial escape conic — a visible map/TS gap at the launch->escape handoff, with the icon jumping across it at warp. `PadAlignLaunch` bails for this regime, so it was left unaligned.

## Approach (render/playback-time only; recorded data immutable; targeting byte-identical throughout)

The PR's first commit (forward launch-hold) was **superseded** — it inserted the hold as in-cycle dead time, which overflowed the back-to-back cadence budget (a clamp truncated it, reopening the seam) and created a pre-launch pad wait. Both failed playtest. The working approach:

1. **Borrow-at-launch / repay-at-SOI-exit.** Launch each loop `delta_N` earlier (`delta_N = Off_N mod T_sid`) so the in-Kerbin-SOI ascent+parking+escape render at the recorded rotation and meet the frozen escape conic (seam closes); repay `delta_N` as a coast hold at the SOI exit so everything downstream (transfer, Duna arrival + its hold, landing) is byte-identical to baseline. **Validated in-game** (aligned loops collapse `residualDeg` to 0).
2. **Warp-to fix.** The Missions "Warp to..." / next-launch now targets the actual launch instant; the single 15s lead comes from `ApplyJumpLead` (was double-applied). **Validated in-game.**
3. **Seam-bridge fix.** Once the launch meets the conic, the near-meet seam bridge (a 74-deg conic merge slice) became a spurious bulging extra segment; `BridgeMinAngleRadians` raised to 5 deg so a leg that already meets the conic skips the bridge. Genuine 5-45 deg gaps still bridge. **Validated in-game.**
4. **Boundary-overlap.** On ZERO-SLACK loops (the per-loop Duna arrival hold fills the whole loop) the launch advance capped to 0 and the seam stayed open (`residualDeg` up to ~211 on loop 19). Fix (Design B, dual-clock): on loops that need it (`delta_N > slack_{N-1}`), render the early-launching next instance as a SECONDARY ghost during the `<= T_sid` borrow window so the launch realigns without the slack cap. The primary instance is far downstream near Duna by then — two ghosts at different places. Already-aligned loops and the targeting path stay byte-identical (the gate guarantees no secondary). The secondary reuses the `overlapGhosts` / `overlapInstanceVessels` storage but never the self-overlap expiry/camera path; it is audio-muted and never camera-bound. **Validated in-game** (zero-slack loops align, brief two-ghost overlap, no camera disturbance).
5. **Secondary ascent icon + review hardening.** The boundary-overlap secondary's proto map presence (icon + escape conic) needs an `OrbitSegment` to exist — and atmospheric ascent has none — so for the first minutes after launch its ascent line drew with **no icon** riding it (map; TS only looked clean because warp skipped the gap). Fixed with an additive non-proto marker riding the shared polyline on both surfaces (`TryResolveBoundaryOverlapSecondaryMarker`), self-gating once the escape Segment's proto icon takes over. Plus four fixes from a clean-context adversarial review (no criticals; clock / watch isolation / audio mute / map lifecycle / targeting byte-identity verified solid): **H1** forward-arc cache was recording-id-keyed, so the primary + secondary forward passes collided in one frame and the primary's forward conic vanished → namespaced the secondary under `ForwardArcDictKey`; **H2** the engine's secondary 3D mesh was torn down by the `decision != Render` early-return in a two-recording chain launch → hoisted the decision and drive the secondary when the primary is hidden; **M1** `GetGhostVesselPidForRecording` fell through to the secondary's pid for a renderHidden primary → `boundaryOverlapSecondaryPids` exclusion set; **L1** corrected a stale comment (the non-owned secondary leg's `ghostPid` is inert). **Validated in-game.**
6. **Small-gap chord bridge.** With alignment engaged the ascent->escape seam collapses to a few degrees, but a small visible gap remained — the forward-drawn ascent end LEADS the icon at the live rotation and lands a few degrees short of the inertial conic, the gap closing as the icon climbs. The merge-slice bridge is muted below 5 deg (its fixed ~200-370 km off-chord bulge dwarfs a small gap), so it was left unfilled. Added a STRAIGHT-CHORD band: `ClassifySeamBridgeByAngle` routes `(0.5 deg, 5 deg]` to a zero-bulge straight chord (`TryBuildSeamBridgeChordLocalPoints`, leg end -> conic near endpoint, no conic clip) that closes the gap and re-fits per frame; `(5 deg, 45 deg]` still the merge slice, `>45 deg` / `<=0.5 deg` still skip. **Validated in-game** (`src=chord`, leg=0 side=end, angleDeg 0.7-2.9).

## Tests

`dotnet test`: **15831 passing, 0 failing.** New across the cycle: `BoundaryOverlapClockTests` (incl. byte-identity + targeting fences + the `SecondaryMarker_*` gate), `BoundaryOverlapLoggingTests`, `GhostTrajectoryPolylineBuildTests.ForwardArcDictKey_*` (H1) + `SeamBridgeChord_*` + `ClassifySeamBridgeByAngle_FourBands`, and in-game `GhostLifecycle` tests (`BoundaryOverlapSecondary_*`: muted/flagged/unwatched, at-most-one, map-instance, marker-resolves-during-ascent-gap, engine-mesh-survives-hidden-primary, does-not-shadow-recording-pid); plus the borrow-repay / warp / bridge suites from the earlier commits.

## Status

**All six parts are playtest-validated in-game (map + TS).** Kept as a draft pending review — ready to mark for review.

Deferred (a larger scope decision, not needed here): the architecturally cleanest launch-seam fix would de-rotate the forward-drawn ascent / extend the re-aim launch-rotation alignment to FAITHFUL looped launches (Duna One already closes this seam that way); the chord is the minimal render-only close for the residual the alignment leaves.

Design: `docs/dev/design-reaim-launch-hold-seam.md` + `docs/dev/plan-launch-boundary-overlap.md`.


vl3c:
Playtest on s15 / Kerbal X #2 FAILED; do not merge. Deployed DLL reverted to baseline (main 8e0413725).

Two failures observed (log: logs/2026-06-17_2021_launch-hold-fail):
1. Launch happened ~HN (cycle 13: HN=5419s, ~1.5h) AFTER the navigable launch window, with the ghost absent in between. The navigable/displayed launch window must BE the aligned time L_N + H_N, not L_N with the ghost held absent after it.
2. The launch->escape seam did NOT close after launch. Prime suspects: (a) spanStart != first body-fixed ascent sample for aa48920e (pad-sit idle frames -> aligning spanStart aligns the wrong instant); (b) the polyline LINE is drawn at the live planet rotation and is not actually derotated by the span-clock shift the icon uses, so only the icon aligns and the line drifts. Needs an instrumented measurement of the seam angle at the launch handoff before any re-attempt; not a blind re-fix.

PR #1188, opened 2026-06-23
Repair 8 stale/brittle in-game test failures (no mod regressions)

vl3c:
## Summary

The full in-game test suite (Ctrl+Shift+T) reported **8 FLIGHT/TRACKSTATION failures**. Investigation (collect-logs snapshot `2026-06-23_2035`) shows **none is a mod/production bug** — all 8 are test-side issues: stale premises left behind by recent *correct* production changes, plus four "isolated-run only" staging tests that need a hand-built craft and should skip gracefully.

This PR repairs the tests so the suite reflects the current (correct) contracts. **Test-only — no production code changed.**

## The 8 failures and fixes

| Test | Root cause | Fix |
|---|---|---|
| `ReStockCompat.TuningStanddownEngagesForCoveredTunedParts` | v1 `microEngine` IS ReStock-covered now — `ReStock/PatchesLegacy/…/restock-engines-liquid-0625.cfg` authors a fresh `EFFECTS{fx-ant-running}`, and the index scans `PatchesLegacy` by design. The negative assertion was stale. | Move `microEngine` into `coveredTunedParts`; assert selectivity on `mk1pod.v2` (ReStock-patched mesh, no plume EFFECTS). |
| `GhostTrajectoryPolyline_MarkerHoldsLastGoodAcrossDropout` | `94599731f` (descent-icon, `ResolveUndrawnLegFallback`) changed the hold to fire **only for conic-anchored legs**. The test built a non-anchored leg via `RefreshForRecording`, which now correctly returns `FallbackNonAnchoredUseHead`. | Seed a conic-anchored leg via `SetLegForTesting(…, wasAnchored: true, …)`. |
| `ControlledChildBreakup_StampsParentAnchorContract` | Read `RecordingStore.PendingTree`, which is **null during steady FLIGHT recording** (it only holds a tree stashed across a scene change). The live tree is the active tree. | Read `ParsekFlight.ActiveTreeForSerialization` (both the precondition and `FindFirstControlledChildRecording`). |
| `CaptureRPOnStaging`, `SavePathRootThenMove`, `WarpZeroedDuringSave` | Ran on a vessel whose next stage doesn't split into two controllable halves (the sibling `…Seed` test correctly **skipped** for the same reason); these lacked that guard and timed out → failed. | **Skip** when no second controllable vessel appears after staging; **keep failing** only when a controllable split happened but no RewindPoint was captured (the real-regression case). |
| `LoopGhostCreationFrame.FreshLoopGhostIcon` | Premise ("the in-create drive doesn't refresh CoMD") was invalidated by `230a9c346` (Prefix `v.CoMD` write); `159145c62` documents the resulting 25–135° loop-shift offset as *correct*. The proto epoch-bake itself is intact. | Conditional skip-guard with documented reason; bake still proven by `registeredShift` + `DirectorDriveEpochBakePlacesIconOnRecordedPhase`. |
| `LoopBoundaryIconWarpLag.StateVectorReseed` | The isolated test calls a bare `CalculatePhysicsStats()` on a Prefix-owned packed ghost and doesn't reconstruct the production reseed→`updateFromParameters`→stats ordering, so it reads as a no-op. Production seam intact (`GhostMapPresence.cs`). | Conditional skip-guard (when the re-snap is a complete no-op); covered by `LoopIconWarpLagTests` xUnit + the passing sibling `IconDrivePrefix_ResnapsCoMDOntoDriverPos`. |

## Validation

- `dotnet build` (Parsek): **0 warnings, 0 errors**; no KSP deploy.
- `dotnet test`: **16217 passed, 0 failed**.
- The in-game tests can't run headless — a `Ctrl+Shift+T` re-run on the right scenario is needed to confirm the 8 now pass/skip. The 4 staging tests need an isolated FLIGHT session with a 2-pod+decoupler craft to actually exercise (otherwise they now skip cleanly).


vl3c:
## Adversarial review pass (multi-agent)

Ran a 6-lens review (one per fix-cluster + a code-quality lens), each pipelined into an adversarial verifier that tried to *refute* every material finding against the real code. Result: **6 of 8 fixes verified clean**, 2 material findings on the loop guards, 2 reviewer claims refuted.

**Acted on (1):** `FreshLoopGhostIcon` skip-guard `if (angleToRecordedDeg >= 5) Skip` **masked the raw-epoch regression the test guards** — a regressed icon sits at the raw-epoch-at-live phase, which is also ≥5° off recorded, so it would skip instead of fail. Tightened to require the icon be off **both** recorded *and* raw-at-live (the isolated drive-phase artifact is far from both; a regression sits ~0° from raw, so `IsLessThan` still fails). Commit `ab463e3`.

**Verified, no change:**
- `StateVectorReseed` guard — confirmed it does **not** mask a regression (production seam unchanged + xUnit-covered + hard-asserting sibling `IconDrivePrefix_ResnapsCoMDOntoDriverPos`). Kept.
- `mk1pod.v2` negative assertion — one reviewer claimed ReStock authors EFFECTS for it; the verifier read `Patches/Command/restock-command-pods.cfg` and showed those EFFECTS belong to the **distinct `mk1-3pod`** part. The assertion is correct.
- Accessor (#7), marker (#8), staging (#4–6) — all verified clean against production (`ActiveTreeForSerialization`, `ResolveUndrawnLegFallback`/`TryHoldLastGood`, `IsMultiControllableSplit`).

Parsek rebuilds 0 errors; xUnit still 16217/0.

PR #1215, opened 2026-06-27
Remove remaining PendingScienceSubjects blanket-clear footguns (dead ClearRecording + CommitTree duplicate-skip)

vl3c:
Follow-up to #1212 (which scoped the blanket `PendingScienceSubjects.Clear()` in the two non-rewind discard cores). A clean-context review of #1212 surfaced two more instances of the same footgun outside that PR's scope. This PR removes both.

`PendingScienceSubjects` is a static, per-recordingId-tagged list of science collected while a recording is live. A blanket `Clear()` next to a scoped consumer drops a **different** live recording's still-uncommitted science (ledger then under-credits what KSP already banked).

## 1. Dead `ClearRecording` stub (commit 1)

`ParsekFlight.ClearRecording()` and its sole caller `RecordingsTableUI.ShowClearRecordingConfirmation()` are **unreachable dead code** — a repo-wide grep finds zero callers of `ShowClearRecordingConfirmation`, so the "Clear Current Recording" button is no longer wired. The method is also broken even if reached: it only nulls the recorder, never discards the active tree, and blanket-clears `PendingScienceSubjects`. Removed both, plus the stale In-Game Controls doc line. No behavior change (dead code).

## 2. `CommitTree` duplicate-skip blanket clear (commit 2)

`CommitTree`'s two duplicate-skip early returns (`RecordingStore.cs:975`, `:1030` — fired on a redundant re-commit of an already-committed tree id) called `PendingScienceSubjects.Clear()` before returning. That wipes a different live recording's uncommitted science. It is also redundant: **every** commit caller runs the scoped `LedgerOrchestrator.NotifyLedgerTreeCommitted` immediately after `CommitTree`/`CommitPendingTree`:
- `ParsekFlight.CommitTreeFlight` (`:12929`)
- `MergeDialog.Commit` via `CommitPendingTree` (`:91`)
- the `CommitPendingTreeAsApplied` auto-commit seam (documented contract: callers chain `NotifyLedgerTreeCommitted` themselves)

That scoped consumer routes pending science per-recording and preserves unrelated subjects. So the duplicate-skip clears are dropped entirely (a no-op skip should not touch the global pending list); removal can only improve over the old wipe (worst case is a benign leak the next commit routes, never data loss). This also makes the duplicate-skip path consistent with the success path, which never clears.

## Tests

- New `TreeCommitTests.CommitTree_DuplicateSkip_PreservesOtherRecordingUncommittedScience`, **verified to FAIL against the pre-fix blanket clear** and pass after.
- Full suite **16270 green**; the dead-code removal breaks no build/test reference.

No CHANGELOG entry: no user-facing behavior change (dead code + a degenerate-path internal hardening; the 0.10.2 discard-economy line already covers the principle).

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -210,7 +210,7 @@ When a worktree's branch finishes its work, land it via a GitHub PR, NOT a local
 
 - **Toolbar button** - Toggle Parsek UI
 - **Ctrl+Shift+T** - Toggle in-game test runner (works in any scene). Results auto-export to `parsek-test-results.txt` in KSP root.
-- UI buttons: Start/Stop Recording, Preview Playback, Stop Preview, Clear Current Recording
+- UI buttons: Start/Stop Recording, Preview Playback, Stop Preview
 
 ## Debug
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -159,7 +159,7 @@ Pick `<target>` carefully: for new work from main, use `origin/main` because loc
 
 - **Toolbar button** - Toggle Parsek UI
 - **Ctrl+Shift+T** - Toggle in-game test runner (works in any scene). Results auto-export to `parsek-test-results.txt` in KSP root.
-- UI buttons: Start/Stop Recording, Preview Playback, Stop Preview, Clear Current Recording
+- UI buttons: Start/Stop Recording, Preview Playback, Stop Preview
 
 ## Debug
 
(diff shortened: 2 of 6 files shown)


vl3c:
Superseded. Split per request: the dead-`ClearRecording` removal moved to a focused PR #1217, and the `CommitTree` duplicate-skip blanket-clear fix is handled on its own branch `fix-committree-dup-pending-science` (to avoid duplicating that parallel effort). Closing this combined PR.

PR #1265, opened 2026-07-08
test(ingame): LogisticsHarvest capture gates self-discard the auto-record session instead of skipping

vl3c:
## Problem

`LogisticsHarvestRuntimeTests` shares one precondition helper, `RequireIdleFlightWithVessel`, across its three capture gates. It skipped on a live recording:

```
A recording is already active; this test starts and discards its own
A live recording tree already exists; this test creates and discards its own tree and must not touch player data
```

Flight auto-records, so an active recording plus its tree is the NORMAL state of any ordinary session, not an operator error. The precondition was therefore never satisfiable and all three capture gates skipped in every ordinary session, which violates the ALL-TESTS-AUTO self-setup policy.

## Fix

Setup now performs the self-discard shipped in `GrappleCaptureInGameTest` (commit `4f859994c`, branch `logistics-claw-producer`) rather than skipping. `RequireIdleFlightWithVessel` keeps its real skips (no `ParsekFlight`, no active vessel, vessel not loaded+unpacked) and then calls a new `DiscardSessionRecordingForSelfSetup`, which:

- stops the recorder when there is no active tree;
- otherwise discards the session tree through the non-public `ParsekFlight.DiscardActiveTreeForSuppressedSceneExit`, the same reflection surface `RuntimeTests.DiscardActiveTreeForRuntimeTest` uses, failing loudly on `TargetInvocationException` and skipping if the reflection surface ever disappears;
- logs the transition (`wasRecording`, tree id);
- asserts BOTH `IsRecording == false` and `ActiveTreeForSerialization == null` before returning, so the gate never proceeds on a half-cleared state.

Post-discard state (`recorder == null`, `activeTree == null`) is exactly the idle state the tests were already written against, and their `StartRecording` then builds a fresh tree as before.

### Why not `AutoDiscardIdleActiveTree`

That method is `internal` and needs no reflection, and this file's own teardown already calls it, so it is the obvious-looking choice. It is the wrong one for setup: `AutoDiscardActiveTreeCore` calls `LedgerOrchestrator.PreserveIrreversibleLiveGameplayOnDiscard` (re-homes the player's contracts / science / milestones onto the direct ledger), runs `RecalculateAndPatchForCurrentTimelineIfFutureActions`, and shows a "Recording discarded - idle on pad" toast. Applying that to the player's real session recording is exactly the "must not touch player data" hazard the old skip was guarding. `DiscardActiveTreeForSuppressedSceneExit` is the pure in-memory teardown: force-stop the recorder, drop the background recorder without persisting, null the tree. It dereferences the tree unconditionally, hence the null-tree guard around it.

## Tier check (requested)

No promotion needed. All four tests in the class are ALREADY isolated-tier:

```
AllowBatchExecution = false,
RestoreBatchFlightBaselineAfterExecution = true,
BatchSkipReason = IsolatedOnlyBatchSkipReason,
```

So the post-test baseline quickload restores the pre-test world and the discarded session recording costs the player nothing. Discarding the session recording therefore requires no tier change. Only the three capture gates route through the helper; the fourth test (`HarvestRoute_AnalyzesEligible_FromSyntheticRecording`) is read-only over the injected synthetic tree and never took the precondition.

Auto-record cannot silently re-arm after the discard either: every `StartRecording` trigger in `ParsekFlight` is event-driven (launch situation change, first staging on pad, post-switch, deferred EVA), none is a per-frame "not recording, so start" loop, and these tests neither stage nor switch vessels.

## Validation

- `dotnet build Source/Parsek` succeeds, 0 errors, no new warnings (the one `CS0414` is pre-existing in `InGameTestRunner.cs`).
- Full xUnit suite green: **17197 passed, 0 failed, 1 skipped** (the pre-existing intentional `RouteRecoveryLandingHarnessTests.UtMappedIdeal_...` skip).
- Test-only change; no production code touched. Per the repo's review policy, self-reviewed rather than given a full review pass.

The in-game gates themselves still need one operator run (Ctrl+Shift+T, Run All + Isolated, in any FLIGHT scene with a loaded and unpacked vessel) to confirm they now execute instead of skipping. That is the point of the change and cannot be verified headlessly.

## Docs

- `CHANGELOG.md`: one line under Unreleased / Internals & Tests.
- `docs/dev/todo-and-known-bugs.md`: the M2 harvest manual-playtest entry claimed "in-game twins exist ... isolated-run"; it now records that they were unrunnable, what the fix was, and that no pre-run operator action is needed.

## Overlap

None with the concurrent session fixing `EvaSpawnWalkbackOnOverlap` and `ExplosionAnchorPosition_BelowTerrain` (different files). Independent of the unmerged `logistics-claw-producer` branch: this branch is cut from `origin/main`, which does not yet contain `GrappleCaptureInGameTest`, so the pattern is duplicated rather than shared. Once #1262 lands, promoting `RuntimeTests.DiscardActiveTreeForRuntimeTest` from `private` to `internal` would let all three call sites share one helper.


vl3c:
## Review follow-up (multi-agent review + fixes)

Ran a multi-agent review focused on the two questions raised: can these gates run automated, and do they save/restore state correctly. It surfaced a genuine data-loss hole and some inaccuracies. Pushed `52cae0357` with three fixes.

### 1. Data loss via the row play button (was a real blocker)

`RunSingle` (the row play button the `BatchSkipReason` recommends) captured a batch baseline but, unlike `RunAllIncludingFlightRestore`, never called `PrepareBatchFlightRestoreExecution`. `CaptureBatchBaseline` swallows a capture failure (quicksave fails / sidecar locked / disk issue), nulls the baseline, downgrades the mode, and returns normally. With a null baseline the prime is skipped, the post-test restore early-returns, and the final restore is gated off, so a restore-after-run test ran its destructive setup and **nothing ever restored** the player's live recording.

Fix: `RunSingle` now applies the same baseline-availability gate as the Run All + Isolated path, so the test skips instead of mutating live flight state with no baseline. The happy path was already safe (the review confirmed the baseline quicksave persists the active tree via `SaveActiveTreeIfAny` and `OnLoad` resurrects it) - this only closes the swallowed-capture-failure branch.

### 2. Discard reordered to be non-destructive

The discard used to live inside the shared precondition helper, which ran **before** each test's converter/drill capability check, so a plain stock craft had its recording discarded and then the test skipped for a run that could never happen. The helper (now `RequireFlightWithUnpackedVessel`) no longer touches the recording; each test self-discards as the first statement of its recording body, **after** the capability check and **before** the log observer. So a stock craft skips without touching the player's recording, and the live recorder's teardown never pollutes the captured log stream the assertions read (moving the discard after the observer would have; that was caught and avoided).

### 3. Consistent teardown + honest docs

- `CleanupRecordingAndTree` now tears the test's own tree down through the same pure suppressed-scene-exit path as setup, not `AutoDiscardIdleActiveTree` (whose ledger recalc + "idle on pad" toast are wrong for synthetic test data).
- Removed the doc/comment references to a `GrappleCaptureInGameTest` that does not exist on this branch (`main` does not carry it); the mirrored surface is `RuntimeTests.DiscardActiveTreeForRuntimeTest`.
- Corrected the CHANGELOG/todo overstatement: these gates are not run-on-any-craft. They need a `BaseConverter` (toggle/warp), an activated `BaseDrill` on ore (D5 catch-up), or the injected `tree-drill-harvest-m2` (analysis), and skip loudly otherwise.

### Tier check

Unchanged from the original PR: all four tests remain isolated-tier (`AllowBatchExecution=false`, `RestoreBatchFlightBaselineAfterExecution=true`); no promotion needed.

### Validation

Build clean; full xUnit suite green (17197 passed, 0 failed, 1 pre-existing skip). The in-game gates still need one operator run (Ctrl+Shift+T -> Run All + Isolated, on a converter/drill craft) to confirm they execute; that cannot be verified headlessly.

PR #1294, opened 2026-07-11
Fix loaded-vs-unloaded parity in supply-route inventory delivery

vl3c:
## Summary

Fixes three parity gaps in the supply-route inventory delivery path (probe / planner / writers), verified against decompiled stock (KSP 1.12.5):

- **Gap A - stacked quantity lost on the loaded path:** the manifest compresses identical stored parts per identity hash with the delivered count in the STOREDPART wrapper's `quantity`, but the loaded writer rebuilt a `ProtoPartSnapshot` from the inner PART node and stock `StoreCargoPartAtSlot` unconditionally stores `quantity = 1`. A Quantity=10 item delivered 10 units unloaded but 1 unit loaded. The loaded writer now raises the slot's stack via stock `UpdateStackAmountAtSlot` (clamped to the snapshot-derived `stackCapacity`) and reads the slot back for the actual, which also covers stock's true-on-no-op return.
- **Gap B - no volume/mass admission:** decompilation shows stock's packed-volume/mass enforcement lives in the storage UI, not in `StoreCargoPartAtSlot`, so BOTH automated branches bypassed it. The probe now runs an explicit admission (`ProbeInventoryUnitsThatFit` + `ConsumeInventoryCapacity`, pure core `ComputeUnitsThatFit`) mirroring stock's accounting: prefab `packedVolume` and prefab mass per unit against the container PREFAB's `packedVolumeLimit`/`massLimit` on both branches (limit <= 0 unlimited, `packedVolume < 0` or unresolvable item/container prefab fails closed, 1e-9 unit epsilon so an exact fit is not floored short). The planner marks non-fitting units skipped, so the writers never receive them.
- **Gap C - non-stackable over-compression:** the planner assigned ONE slot per manifest item regardless of Quantity, persisting invalid stock state (quantity=10 in a stack-1 slot) unloaded. The planner now splits each item into ceil(Quantity / stackCapacity) slot-sized stacks (`InventoryDeliveryLine.Units`), resolving stackCapacity from the STOREDPART wrapper, then the inner PART's `moduleCargoStackableQuantity`, with the live prefab fallback only for snapshot-less items and stack-1 forced for resource-bearing payloads.

Delivery actuals are now unit-accurate: `ReadInventoryActualCount` sums stored units (was: line count), the `InventoryWriter` seam carries `(item, slot, units)`, and the delivery summary logs `inventoryUnits=actual/attempted inventoryUnitsSkipped=N`. The container budget and per-part footprints are memoized per planning pass.

## Testing

- 21 new xUnit tests: planner multi-slot splitting / volume rejection / stack-capacity resolution (`RouteDeliveryPlannerTests`), admission core + unloaded STOREDPART builder (`InventoryDeliveryParityTests`), unit-accurate apply + skip logging (`RouteOrchestratorDeliveryTests`), stack-quantity codec round-trip (`RouteCodecTests`). Full suite: 17550 passed, 0 failed.
- New in-game test `Delivery_LoadedVessel_StacksInventoryQuantityIntoSlot` (Logistics category, isolated-run): delivers a units>1 stackable payload to the loaded active vessel via the production writer and asserts the slot's stored stack amount and the unit-accurate actual. PENDING-OPERATOR: needs a FLIGHT vessel with an inventory container holding a stackable cargo part (e.g. an EVA Repair Kit) plus one empty slot.
- Second commit is the pre-PR review pass (branch-symmetric container limits, fail-closed edges, exact-fit epsilon, budget memoization).

## Notes

- Complementary to any follow-up that extends the probe/planner/writers to scan ALL inventory modules: this PR changes HOW MUCH and WHETHER items fit; the first-inventory-module scope is unchanged, and the loaded budget now routes through `LiveDeliveryWriters.FindFirstInventoryModule` so that scope has one definition to extend.


vl3c:
Added a third commit closing the main coverage gap from the self-review: `Delivery_Probe_AdmissionMatchesIndependentBudget`, a read-only, batch-safe in-game test that cross-checks the live probe's admission wiring (budget walk, prefab-limit source, occupancy accounting, consumption tracking, fail-closed unknown prefab, stackable-quantity read) against an independently computed budget run through the xUnit-pinned pure core. Unlike the stack-store test it needs no special setup beyond an inventory container on the FLIGHT vessel, so it runs in every ordinary batch.

PR #1308, opened 2026-07-12
M-A6.1: provisioner live execution (CLONE/BUILD-TT/INSTALL/VERIFY/repair) + all pins resolved

vl3c:
## Summary

Completes the provisioner: the live phases replace the abort-loudly stubs, and all three stack pins are resolved with dual-source verification.

- **CLONE**: instance creation under automation/<profile> with the .provision-incomplete marker (written first, cleared last), free-space precheck, extended-length path handling, junctions for the stock payload trees, and instance-side content verification of every dev-sourced mod at copy time.
- **BUILD-TT**: TestingTools compiled from the two pinned v0.5.4 sources exported via git show at the peel-verified commit (never a working tree), SDK shim csproj against the dev KSP DLLs + release kRPC assemblies, metadata assertions that AutoLoadGame is ABSENT and all six capability RPCs are PRESENT in the built assembly.
- **INSTALL**: the three pinned artifacts downloaded fresh each run with sha256 enforcement (mismatch aborts printing both hashes), zip-slip-guarded extraction, per-DLL hashes into the manifest (hlib admission compatible).
- **VERIFY/--repair**: instance-side re-hash of every component INCLUDING dev-sourced mods (review blocker fix), junction realpath probes, settings + buildID64 shas; repair converges only drifted components, with scoped deletion strictly inside the instance GameData so injected files are removed.
- **Safety hardening from review**: EC-16 aliasing re-checked on realpath-RESOLVED paths before any write (defeats pre-existing junction aliasing), reparse points skipped in all copy/hash walks, log buffering until the gate passes, clean EC aborts with lock release on dotnet-missing/OSError paths, DEPLOY aborts without an unambiguous Parsek.dll source (no sibling-worktree fallback).
- **Pins**: kRPC 0.5.4 zip and KRPC.MechJeb 0.7.1 release hashed from official GitHub assets; MechJeb2 2.15.1.0 pinned via the CKAN-meta record (fresh-download sha256 matches CKAN's independent hash; archive.org mirror documented as rot fallback).

Idempotency short-circuits and per-component installed-file inventories are tracked M-A6.2 follow-ups.

## Review

Clean-context adversarial review: destructive-path safety verdict 'fenced' (no scenario damages anything outside automation/ + the cache); 1 blocker (source-side-only dev-mod verification) and 7 hardening should-fixes, all applied.

## Test plan

harness/provision: 123 unittest green (was 70 pre-M-A6.1); harness/lib: 131 unaffected; both profiles dry-run exit 0; BUILD-TT validated for real against mods/krpc@v0.5.4 (compiled, S-4 + capability assertions passed). The real instance-creating smoke run is the coordinated next step after merge.


vl3c:
Live-smoke addendum: after the branch review, the real provisioning smoke + first unattended harness run were executed from this branch and surfaced seven real-world fixes, all committed here:

- mklink argument separators (cmd parses forward slashes as switches)
- unconditional junction pre-clear + lstat-based link probes (os.path.exists/isdir/islink can all report False for a healthy junction, reparse tag 0xA0000003)
- layout-aware component-DLL verify paths (MechJeb2 ships under Plugins/)
- kRPC PluginData settings stamped for unattended operation (autoStartServers/autoAcceptConnections=True, confirmRemoveClient=False) with a krpcSettingsSha256 manifest+VERIFY drift check
- seam LoadGame two-phase completion now detects failed loads (return-to-menu -> ERROR load-failed-returned-to-menu; 300s budget -> ERROR load-timeout) instead of hanging PENDING - found when an incompatible modded save NRE'd FlightDriver.Start
- MechJeb2 pin resolved via CKAN-meta (dual-source sha256)

End state on this branch: the stock-minimal instance provisions to VERIFY drift=0 exit=0, and the first unattended harness run drove real KSP end to end through the seam and correctly classified the incompatible-fixture failure. Suites: 18163 xUnit / 142+131 python, all green.

PR #1309, opened 2026-07-12
Analyzer follow-ups: INV9 provisional-severity split, unparsable-save test, dump relocation

vl3c:
## Summary

The three tracked follow-ups from the retroactive review of the analyzer tuning:

- **INV9 severity split by MergeState**: a CommittedProvisional (still-rewindable) recording whose own rewind save file is missing now FAILs with the distinct token missing-rewind-save-provisional; Immutable stays WARN. This restores the one detection class the earlier WARN downgrade blurred. Verified against the five baselined historical saves: all stay GREEN (every existing dangling case is Immutable), so no baseline refreshes were needed and the severity-escalation guard was not tripped.
- **Unparsable-rewind-save FAIL path test**: the last untested file-state FAIL in INV9 now has a deterministic fixture (whitespace-only .sfs, which ConfigNode.Load reliably nulls).
- **RecordingSectionDump relocation**: the triage tool now writes to PARSEK_DUMP_RESULTS / PARSEK_ANALYZER_RESULTS / the test output dir - never inside the triaged save - making it literally read-only over saves.

## Test plan

Full suite 18167 passed / 0 failed; analyze-historical-saves.ps1 green before and after. A targeted clean-context review of the verdict-behavior change is queued before merge (will comment here).


vl3c:
Clean-context targeted review (F1 verdict-behavior focus). Verdict: the new FAIL class is safe to ship against current data (all suite tests green, baselined saves unaffected), but the F1 rationale cites the wrong production contract in four places and the MergeState gate is a heuristic with a narrow benign false-FAIL path. Text fixes requested; no verdict-logic change strictly required.

Findings:

- MEDIUM (rationale misattribution, 4 places): the F1 justification says a missing save means "a real Rewind-to-Separation would fail to find its quicksave". Wrong flow. Recording.RewindSaveFileName (parsek_rw_*, Parsek/Saves) drives Rewind-to-LAUNCH: RecordingStore.InitiateRewind / BeginRewindForOwner logs "Rewind-to-Launch initiated ... this is NOT a Re-Fly (no RewindPoint / ReFlySessionMarker / MergeJournal)". Re-fly of a CommittedProvisional UF slot resolves through rp_* RewindPoints (RecordingsTableUI.ResolveUnfinishedFlightRewindRoute -> TryResolveRewindPointForRecording), the system the rule header itself scopes OUT of INV9. The pre-existing header line 13 ("the Rewind-to-Separation quicksave captured at recording start") had the same error; F1 propagates it into the new class comment, the test comment, the design doc INV9 section, and the todo entry. Please s/Rewind-to-Separation/Rewind-to-Launch/ in all four; the design doc is the binding authority.

- MEDIUM (gate is a heuristic, not a production contract; narrow false-FAIL path): production rewindability of the parsek_rw_ save is MergeState-agnostic. GetRewindRecording / CanRewind / InitiateRewind never read MergeState, so "a sealed (Immutable) recording can no longer be rewound at all" is false for this artifact: an Immutable root with the file present still shows an enabled R button, and its dangler breaks the same user affordance the provisional case does. Symmetrically, the benign-dangler mechanism the WARN class excuses (DeleteRecordingFiles removing a shared parsek_rw_ file without reference-counting; CopyRewindSaveToRoot leaves the same filename on both the capturing segment row (ParsekFlight.cs:4604) and the root) is also MergeState-agnostic, so a CommittedProvisional root can acquire a dangler by the exact benign path -> false FAIL. Empirically rare (all 8 historical danglers are Immutable) and recoverable via a baselined entry with a human reason, so the split can stand as a pragmatic escalation, but the comment should say what is actually true: an open provisional slot is recent, still-active state where a dangling launch save is far more likely a live bug than archaeology, not "provisional = rewindable / immutable = not".

- LOW (F3 left one doc line dangling): docs/dev/design-autotest-offline-analyzer.md ~line 243 still says the section dump lands at `<save>/analysis/<id|all>.sectiondump.txt`. F3 moved it to the results-dir convention; update this line to match the new ResolveResultsDir() precedence.

Verified clean:

- NotCommitted rows: they can appear in the offline model (zombie re-fly provisionals persisted mid-refly; LoadTimeSweep discards them on next load) and the rule routes them to the WARN else-branch, which is right. Hydration cannot manufacture a false FAIL: Recording.MergeState field-defaults to Immutable and RecordingTreeRecordCodec only writes mergeState when != Immutable, so a partially loaded row lands in the WARN class.
- CitedContract: RecordingPaths.BuildRewindSaveRelativePath on the new FAIL is the right mechanical choice (location contract, matching the WARN), with the id-validator distinction explained.
- F2 determinism: sound. Whitespace-only content gives File.ReadAllLines >= 1 line, so KSP's PreFormatConfig takes the processing path (yields no structure, Load returns null) and never reaches the zero-line Debug.LogError branch that ECall-faults on truly empty files under xUnit. Even in a hostile environment the ParsesAsConfigNode catch -> false produces the same verdict.
- Baseline interplay: confirmed both belts. The baseline key digest differs in letters ("missing-rewind-save" vs "...-provisional"), so an escalated finding never matches its old WARN entry; and BaselineFilter refuses any match whose current level exceeds CapturedLevel (BASELINE-SEVERITY-ESCALATED, BaselineFilter.cs:229). Historical Immutable danglers keep the unchanged WARN token and stay baselined, consistent with the reported all-5-saves-green result.
- Tests: dotnet test --filter Inv9 8/8; full suite 18167 passed, 0 failed, 1 known skip (RouteRecoveryLandingHarnessTests UT-mapped ideal).

vl3c:
Text corrections landed in 358dad315.

MEDIUM (rationale misattribution, 4 places): s/Rewind-to-Separation/Rewind-to-Launch/ applied in the Inv9RewindPoint class/header comments (including the pre-existing header-line misattribution), the test comment, the design doc INV9 section, and the todo entry. RewindSaveFileName (parsek_rw_*, Parsek/Saves) drives Rewind-to-Launch; the rp_* RewindPoint system the header already scopes out is untouched.

MEDIUM (heuristic, not contract): reframed the provisional FAIL split everywhere it is described. The corrected rationale: the MergeState split is a triage-severity heuristic, not a production contract - rewindability of the parsek_rw_ save is MergeState-agnostic (GetRewindRecording / CanRewind / InitiateRewind never read MergeState), so a CommittedProvisional recording is graded FAIL only because it is a recent, still-active slot whose dangling Rewind-to-Launch save is far likelier a live bug than the historical shared-delete residue tolerated on sealed rows. The false 'a sealed recording can no longer be rewound at all' claim is dropped, and the narrow benign false-FAIL path (a provisional root whose shared save a sibling deleted) is noted with baselining as the recovery.

LOW (F3 doc line): design-autotest-offline-analyzer.md ~L243 now points at the relocated output - <results-dir>/<id|all>.sectiondump.txt via ResolveResultsDir() precedence (PARSEK_DUMP_RESULTS, else PARSEK_ANALYZER_RESULTS, else AppContext.BaseDirectory/section-dumps) - instead of the old in-save <save>/analysis path.

Text-only; dotnet test --filter Inv9 green (8/8).

